PR #248, opened 2015-01-05
use rails-assets bower gems rather than checking in code, removes vestig...

mejackreed:
...le gem

We shouldn't check in js if we don't have to. Now using rails-assets service for bower packages. Note that history.js is not correctly packaged in bower (no main files specified), that is why it is not included here.


eliotjordan:
After bumping up the jettywrapper version, all the tests passed for me. However, during startup I get a bucket of css deprecation warnings and the leaflet map styling is broken. Related to issue #249?  Perhaps we should update jettywrapper in a separate PR so other requests can pass their tests?

mejackreed:
@eliotjordan The deprecation warnings come from Rails 4.2. 

Have you regenerated the test app?

I had to update jettywrapper version here to latest, so that tests would pass on new travis environment.

mejackreed:
@eliotjordan See https://travis-ci.org/geoblacklight/geoblacklight/builds/45975320#L187

This is the error you get without updating jettywrapper. Went ahead and did it here and then it should be fine for subsequent pr's.

eliotjordan:
@mejackreed I did regenerate the test app. The warnings disappeared during the tests, however when running the app via rails -s, the styling for the map was obviously broken. Perhaps something's up with my test VM. 
screen shot 2015-01-05 at 4 54 55 pm

mejackreed:
Very strange. It doesn't appear that the leaflet style require is going through for some reason.

mejackreed:
@eliotjordan pulled the jettywrapper change out into its own pr.  https://github.com/geoblacklight/geoblacklight/pull/251

mejackreed:
Depends on #251

mejackreed:
The Leaflet issue is addressed here: https://github.com/geoblacklight/geoblacklight/commit/d59339869c16bafd502506008e12af26fd892cb0 . It appears to be an issue with Rails 4.2 and later Blacklight versions. I can only recreate the issue in Blacklight 5.8.2 with Rails 4.2 . #252 should fix this.

eliotjordan:
@drh-stanford Looks good to me.

PR #250, opened 2015-01-05
adds symlink app/views/catalog to app/views/geoblacklight/catalog

mejackreed:
Enables partials to be called, even if they have already been overridden. 

See https://github.com/projectblacklight/blacklight/pull/725


mejackreed:
Depends on #251

mejackreed:
For future reference. This doesn't work as symlinks do not come through RubyGems.

PR #253, opened 2015-01-07
fixes search bar display on homepage

mejackreed:
After Blacklight 5.8.2 update search bar on home page was displaying incorrectly, this fixes it.


eliotjordan:
All is right with the search box.

PR #255, opened 2015-01-07
add rails-assets as gem source

### LINKED ISSUE #254: Rails Assets breaks install template
See #248. When generating a new GeoBlacklight appication, bundle install cannot find Bower components.

	rails new geoblacklight-app -m https://raw.githubusercontent.com/geoblacklight/geoblacklight/master/template.rb

	...

	Resolving dependencies...
	Could not find gem 'rails-assets-readmore (>= 0) ruby', which is required by gem 'geoblacklight
	(>= 0) ruby', in any of the sources.

eliotjordan:
Need to add rails-assets gem source so bundle install can run without errors. Closes #254

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -2,6 +2,8 @@ gem 'blacklight'
 gem 'geoblacklight', github: 'geoblacklight/geoblacklight'
 gem 'jettywrapper'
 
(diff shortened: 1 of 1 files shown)


mejackreed:
:+1:

PR #256, opened 2015-01-08
move map and attribute table to its own partials

mejackreed:
Pulls out map and attribute tables to new partials. This helps with local customizations.

Also removes symlink created in #250 that apparently does not work.


drh-stanford:
This seems to work for me ok, but my install is a bit iffy.

mejackreed:
You could

```
rake engine_cart:clean && rake engine_cart:generate
```
to reinstall the test app. This pr really just moves all the markup for the respective blocks to separate partials.

PR #265, opened 2015-01-21
adds spatial search constraints to breadcrumb

mejackreed:
screen shot 2015-01-20 at 8 37 40 pm


eliotjordan:
Works great. Passes all tests on my end. 

I wonder, though,  if users who are not 'geo' savvy might be confused by the text in the breadcrumb. Should we expect our users to know what a bounding box is and how to interpret the coordinates? Personally, I like the extra information, but I might not be typical. Perhaps this is something to look at during usability testing.

mejackreed:
@eliotjordan yep thats the plan. One of the things that has already come up is users don't know when/why a spatial search is happening . This should start to help address that. More tests coming up today and next week, so the earlier this gets in the earlier i can test it.

eliotjordan:
@mejackreed Roger that.

PR #267, opened 2015-01-22
adds back in icons to toolbar

mejackreed:
After document actions framework implemented, icons went away. Now they are back!

screen shot 2015-01-22 at 7 17 49 am


eliotjordan:
Icons! Welcome back friends. Passing tests. Looking fabulous.

PR #274, opened 2015-01-29
Constrain search bounds to 360 degrees

### LINKED ISSUE #273: bbox for spatial search at world scale seems to be incorrectly calculating
Doing a world scale spatial search tends to not include results. This appears to be an issue with the way that the spatial search js calculates the bounding box it is looking at.

gravesm:
Fixes #273.


mejackreed:
@gravesm See https://github.com/geoblacklight/geoblacklight/pull/275. #274 and #275 combined fix the issue where a result that Intersects but not IsWithin does not show up.

PR #275, opened 2015-01-29
use boost query rather than q

mejackreed:
This fixes the rest of the problem. They q param to solr wasn't really what we wanted and was causing results to not show up which intersected.


eliotjordan:
+1 IsWithin was too restrictive.

PR #280, opened 2015-02-02
remove gem dependency of rails-assets

mejackreed:
Hopefully once and for all will resolve issues that new installers with rails-assets.


eliotjordan:
Passing tests on my end. I'm able to generate a functional GeoBlacklight instance. Rails-assets is a bit tricky it seems.

mejackreed:
thanks @eliotjordan !

PR #282, opened 2015-02-13
add back in js assets, rails-assets was causing more headaches than it w...

mejackreed:
...as worth


mejackreed:
:bow: @gravesm

PR #288, opened 2015-02-19
adds in caret in results view for toggle

mejackreed:
screen shot 2015-02-18 at 7 00 47 pm

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/results.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/results.scss
@@ -27,3 +27,13 @@
   @extend .text-span;
   @extend .hide-overflow;
 }
+
+.caret-toggle {
+  @extend .fa, .fa-caret-right;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/geoblacklight/modules/results.scss
```diff
@@ -26,4 +26,13 @@
 .index_title {
   @extend .text-span;
   @extend .hide-overflow;
+}
+
+.caret-toggle {
+  @extend .fa, .fa-caret-right;
+  width: 10px;
+
```
gravesm: Do we want to add a ``cursor: pointer`` here? I get the select tool when I mouse over the caret which seems a little weird.

mejackreed: @gravesm great point. I force pushed a change with `cursor:pointer` added.

PR #292, opened 2015-02-23
Add metadata tool

### LINKED ISSUE #75: View metadata from ISO19139 or FGDC source
Instead of downloading the original metadata, we should have an option to view it as HTML.

mejackreed:
Closes #75 

Adds in a metadata tool, with metadata viewing if a reference is designated to be shown.

screen shot 2015-02-23 at 7 12 04 am
screen shot 2015-02-23 at 7 11 58 am


drh-stanford:
What's the status on this? Should we wait for https://github.com/OpenGeoMetadata/GeoCombine/pull/6 and implement the HTML rendered view?

mejackreed:
I think this is mergable now... and https://github.com/OpenGeoMetadata/GeoCombine/pull/6 is an enhancement.

app/views/catalog/_metadata.html.erb
```diff
@@ -0,0 +1,12 @@
+<% document ||= @document %>
+
+<div class='metadata-view'>
+  <% document.references.refs.each do |ref| %>
+    <% if Settings.METADATA_SHOWN.include?(ref.type.to_s) %>
+      <div class='label label-default'>
+        <%= formatted_name_reference(ref.type) %>
+      </div>
+      <%= CodeRay.scan(Geoblacklight::Metadata.new(ref).metadata, :xml).div(css: :class).html_safe %>
```
gravesm: What do you think about adding a tab_width option:

``<%= CodeRay.scan(Geoblacklight::Metadata.new(ref).metadata, :xml).div(css: :class, tab_width: 2).html_safe %>``

The `columbia-columbia-landinfo-global-aet` layer, for example, has 8 spaces for tabs which puts most of the XML off the visible pane requiring a lot more horizontal scrolling.

mejackreed: Great suggestion. Added and force pushed the change up

PR #296, opened 2015-02-24
update solr_bbox to use WKT syntax

drh-stanford:
in preparation for switching to `solr.BBoxField` for `solr_bbox`


drh-stanford:
will resubmit after cleaning out things that read `solr_bbox`

PR #305, opened 2015-03-02
Configure download

mejackreed:
Does a few things here:

 - moves `Download` classes underneath the `Geoblacklight` module. 656c316
 - Allows subclassed downloads to take an options parameter which gets merged in. This allows for a `timeout` parameter that can be configured individually on instantiation independent of an apps configuration


mejackreed:
Just a note this is intended to now allow for the ability to precache downloads that need an extra long timeout

mejackreed:
Also just to be clear, this passed all tests but reduced code coverage (we are not going to unit test private methods).

PR #311, opened 2015-03-16
update to blacklight 5.10.0 and fix deprecation errors

eliotjordan:
Updated the blacklight gem to 5.10.0 and took a swing at fixing the resulting deprecation errors. The Blacklight release notes are somewhat vague on on how to add the new params in SearchBuilder instead of in the CatalogController. If anyone knows a more elegant method, please jump in.


mejackreed:
:+1:

PR #313, opened 2015-03-24
Update bl 5.12.0

mejackreed:


eliotjordan:
Looks good. Only complaints are from the range limit plugin.

PR #325, opened 2015-04-27
Solr doc fetch... and precache downloads

mejackreed:
This pull request adds class methods for finding Solr documents to the `Geoblacklight::SolrDocument` class. Also enables the ability to precache downloads using a rake task.

Rake task call would look something like: 

```sh
$ rake 'geoblacklight:downloads:precache[tufts-cambridgegrid100-04, shapefile, 50]'
```

### DIFF
--- a/lib/geoblacklight.rb
+++ b/lib/geoblacklight.rb
@@ -8,6 +8,7 @@ module Geoblacklight
   require 'geoblacklight/exceptions'
   require 'geoblacklight/view_helper_override'
   require 'geoblacklight/item_viewer'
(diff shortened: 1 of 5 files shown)


eliotjordan:
Everything looks good here. Is the use case that a manager might want to pre-cache popular datasets? Or cache directly after ingest? Seems useful, especially for large files.

mejackreed:
Yep both of those. Another use case it was developed for: if you want to set a lower download timeout for user downloads, but precache those files that will take longer.

PR #327, opened 2015-04-28
use geoblacklight-icons as a gem

mejackreed:
This should make updating the icon font easier from now on. No need to copy over or include the font or SASS... just use the gem


eliotjordan:
Very nice. Much better than copying by hand. :rocket:

PR #330, opened 2015-05-12
Adds documentation and tests for Reference and References

mejackreed:
24a82fa adds comments and specs for Geoblacklight::Reference

b448d91 adds comments, also makes some methods private


eliotjordan:
:thumbsup: I just glanced at the source. I'll take a longer look later this afternoon.

PR #334, opened 2015-05-28
switch from querying solr using rectangle to envelope syntax

### LINKED ISSUE #332: Solr_bbox replacement for map search functionality?
Hi all. I'm working with a development instance of GeoBlacklight, and the search-by-map-position functionality has stopped working for me, I think as a result of updating the app (if that's a possibility?). I see "bbox=" search criteria in the URL that is generated by the app, but it is never successful at faceting by those criteria.

I see from some of the discussion here that bbox has been depreciated –– is the app handling this search in a different way now? I'm running the most recent version of GeoBlacklight (.0.10.0) and Solr 5.0. Any help would be much appreciated!

### LINKED ISSUE #333: Update Solr query syntax to support > Solr 5.x spatial query parsing
Reported by @sgbalogh in #332

mejackreed:
Closes #333 & closes #332 

Preliminary testing with Solr 4.10.4 and Solr 5.1 and spatial search seems to be functional. Holding off on  using Solr 5.x in tests until Blacklight does so. https://github.com/projectblacklight/blacklight/pull/1100

### DIFF
--- a/lib/geoblacklight.rb
+++ b/lib/geoblacklight.rb
@@ -1,6 +1,7 @@
 require "geoblacklight/engine"
 
 module Geoblacklight
+  require 'geoblacklight/bounding_box'
   require 'geoblacklight/catalog_helper_override'
   require 'geoblacklight/config'
   require 'geoblacklight/constants'
--- a/lib/geoblacklight/bounding_box.rb
+++ b/lib/geoblacklight/bounding_box.rb
@@ -1,0 +1,44 @@
+module Geoblacklight
+  ##
+  # Transforms and parses a bounding box for various formats
+  class BoundingBox
+    ##
+    # @param [String, Integer, Float] west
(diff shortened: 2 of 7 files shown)


eliotjordan:
@mejackreed I see your comments at the bottom of this [solr wiki]( https://cwiki.apache.org/confluence/display/solr/Spatial+Search). This a related issue, yes? Solr 5.x only supports spatial queries with the form ```Intersects(ENVELOPE(....))``` ?

mejackreed:
@eliotjordan yes from what I can tell, Solr 5.x only supports the queries with explicit envelope syntax. The documentation feels a bit unclear, but I've tested this branch on solr 4.10.4 and solr 5.1 successfully. 


The current (older) syntax is definitely not supported on Solr 5.1

eliotjordan:
Solr documentation could use some help for sure.

PR #338, opened 2015-06-19
Add JavaScript testing framework Teaspoon

mejackreed:
- enables JavaScript specs (see spec/javascripts/geoblacklight_spec.js)
 - should allow us to start better testing the JavaScript as more complexity is added.
 - uses Jasmine

*Note:* engine_cart:clean && engine_cart:generate will need to be ran

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,12 +8,6 @@ gemspec
 # it'll want sprockets 2.11.0 and we'll have a conflict
 gem 'sprockets', '2.11.0'
 
-# If we don't specify 3.2.15 we'll end up with sass 3.3.2 in the main
-# Gemfile.lock but since sass-rails gets generated (rails new) into the test app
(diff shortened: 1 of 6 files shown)


eliotjordan:
Teaspoon looks nice and seems to play well with GeoBlacklight.

@mejackreed what is/was your reasoning for choosing Jasmine? From what I can tell, both Mocha and Jasmine are similar to RSpec. Jasmine does seem simpler to set up, as Mocha needs additional assertion libraries.

mejackreed:
@eliotjordan Implemented Jasmine because it just seemed easier/simpler to get up and going.

mejackreed:
Also I think it has been kicked around to be added into Blacklight, and the thought was to use Jasmine there as well.

eliotjordan:
The general consensus is that Jasmine is a good way to go, so I'll merge this in.

PR #339, opened 2015-06-19
remove bootstrap-sass gem version locks

mejackreed:
With the latest bootstrap-sass release, we can now remove this from the gem's Gemfile.


eliotjordan:
Nice and easy. :sunrise:

PR #341, opened 2015-07-21
add exports to toolbar, implementing export to CartoDB

mejackreed:
Adds an "Open in CartoDB" link to toolbar for public docs with a wfs web service

screen shot 2015-07-21 at 8 23 15 am

This also opens up the possibility of adding in a "Open in ArcGIS.com" link as well, but there are still some information barriers before adding that feature.

### DIFF
--- a/lib/geoblacklight/solr_document/carto_db.rb
+++ b/lib/geoblacklight/solr_document/carto_db.rb
@@ -1,0 +1,15 @@
+module Geoblacklight
+  module SolrDocument
+    ##
+    # Module for providing external CartoDB download references for a document
+    module CartoDb
+      ##
+      # Returns a url to a file that should be used with CartoDB integration
+      # @return [String]
+      def cartodb_reference
+        return unless public? && download_types.try(:[], :geojson).present?
+        Geoblacklight::GeojsonDownload.new(self).url_with_params
+      end
+    end
+  end
+end
--- a/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
@@ -37,3 +37,8 @@
 .geoblacklight-metadata {
   @extend .fa, .fa-file-text-o;
 }
+
+.geoblacklight-cartodb {
+  @extend .fa, .fa-map-marker;
+}
+
--- a/app/assets/stylesheets/geoblacklight/modules/toolbar.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/toolbar.scss
@@ -27,4 +27,9 @@
     text-align: center;
     width: 1.4em;
   }
+
+  .geoblacklight-cartodb {
+    text-align: center;
+    width: 1.4em;
+  }
 }
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -91,4 +91,17 @@ module GeoblacklightHelper
   def geoblacklight_basemap
     blacklight_config.basemap_provider || 'mapquest'
   end
+
+  ##
+  # Creates a CartoDB OneClick link link, using the configuration link
+  # @param [String] file_link
+  # @return [String]
+  def cartodb_link(file_link)
+    params  = URI.encode_www_form(
+      file: file_link,
+      provider: application_name,
+      logo: Settings.APPLICATION_LOGO_URL
+    )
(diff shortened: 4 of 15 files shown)


eliotjordan:
@mejackreed When I try to load the MA Zip Codes layer (mit-us-ma-e25zcta5dct-2000) I am directed to CartoDB. After logging in (even if already logged in?), the service creates two tables. One, called wfsrequest, is completely empty. Do you experience this?

as0zr0i1vbtsem9jgmytkspe_siwlkg97xpx3uyccfk

mejackreed:
Strange, yes I have the same issue . Not sure how this could be fixed.. Ideas?

eliotjordan:
If you look at the zipped shapefile returned by MIT, they include an extra file called wfsrequest.txt. CartoDB must interpret that as a separate dataset. If you send CartoDB geojson instead, things work as they're supposed to. I still have to login twice, which is weird. It would be great if the oneclick API allowed you to reference a username or the service recognized if you were already logged in. Also, being able to specify a title would be helpful.

lib/geoblacklight/solr_document/carto_db.rb
```diff
@@ -0,0 +1,15 @@
...
+  module SolrDocument
+    ##
+    # Module for providing external CartoDB download references for a document
+    module CartoDb
+      ##
+      # Returns a url to a file that should be used with CartoDB integration
+      # @return [String]
+      def cartodb_reference
+        return unless public? && download_types.try(:[], :shapefile).present?
+        Geoblacklight::ShapefileDownload.new(self).url_with_params
```
eliotjordan: Using geojson output seems to work better:  ````Geoblacklight::GeojsonDownload.new(self).url_with_params````

mejackreed:
Ahh.. ok that all makes sense now. I'll modify it to use GeoJSON. Also I mentioned the login strangeness to @andrewxhill . I think its something on CartoDB's end that needs to be updated.

andrewxhill:
hey @mejackreed regarding the login issue, would you mind tossing a ticket up on github.com/CartoDB/cartodb so we can track it down?

mejackreed:
Sure thing @andrewxhill .. https://github.com/CartoDB/cartodb/issues/4780

juanignaciosl:
FYI CartoDB/cartodb#4780 has fixed the login issue. You'll need to close your session first, but then, after your next login, it will work. Sorry for the inconvenience!

eliotjordan:
@juanignaciosl Thanks for the quick turn around on this issue!

eliotjordan:
@mejackreed I'm getting that wfsrequest with Stanford shapefiles as well, so I think GeoJSON is probably the way to go. I think this is ready to merge after that.

mejackreed:
Yep, I'll get to it in a bit :watch:

eliotjordan:
@mejackreed No hurry. Just giving my official :+1:

mejackreed:
Hey @eliotjordan just another data point:

For shapefile downloads they are much smaller. 

For example (stanford-cz128vq0535):

shapefile: 7.8 Mb
geojson: 25 Mb

This is kind of making me think that we should implement it as the shapefile or make it configurable. What do you think?

eliotjordan:
@mejackreed Yeah. That's a big difference. I could go either way on this. The zipped shapefile is certainly smaller. It looks as though you can [customize this in geoserver](http://docs.geoserver.org/latest/en/user/services/wfs/outputformats.html#shapefile-output-customization) through some kind of templating system. Otherwise that wfsrequest sidecar is included by default. Not sure what the right answer is here.

gravesm:
What's the concern with the size? I'm not really sure what the specifics of the export to CartoDB look like, but I know GeoServer supports HTTP compression.

mejackreed:
@gravesm The roundtrip time of a user waiting for the import to happen. Also there doesn't seem to be a way of giving the file a useful name if its coming from a WFS created GeoJSON file:

https://github.com/CartoDB/cartodb/issues/4828

gravesm:
Ah, gotcha. I guess someone from CartoDB would have to chime in about support for HTTP compression. That still leaves the file name.

mejackreed:
I say :shipit: if no one's opposed... Happy to test the feature out in EarthWorks and see what our users think, we can always change it up later.

gravesm:
:+1:

eliotjordan:
Sounds good to me. If no other objections, I'll merge.

PR #349, opened 2015-10-17
Add webservices layer id enhancement

### LINKED ISSUE #348: Add better user experience for users in "web services" modal
Currently we just display the url to the web services. We have had several reports about the lack of usefulness of this. Proposed adding WxS specific parameters to modal.

mejackreed:
This commit also adds the pattern to start removing hard coded Solr field values from the view using Document Presenter pattern

Closes #348

### DIFF
--- a/lib/generators/geoblacklight/templates/catalog_controller.rb
+++ b/lib/generators/geoblacklight/templates/catalog_controller.rb
@@ -33,7 +33,9 @@ class CatalogController < ApplicationController
 
     config.show.display_type_field = 'format'
 
-
+    # Custom GeoBlacklight fields which currently map to GeoBlacklight-Schema
+    # v0.3.2
+    config.wxs_identifier_field = 'layer_id_s'
 
     # solr fields that will be treated as facets by the blacklight application
     #   The ordering of the field names is the order of the display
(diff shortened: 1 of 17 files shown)


lib/generators/geoblacklight/templates/catalog_controller.rb
```diff
@@ -33,7 +33,9 @@ class CatalogController < ApplicationController
 
     config.show.display_type_field = 'format'
 
-
+    # Custom GeoBlacklight fields which currently map to GeoBlacklight-Schema
+    # v0.3.2
+    config.wxs_identifier_field = 'layer_id_s'
```
eliotjordan: What are the advantages to making the wxs_identifier_field configurable? How would this be used in the wild?

mejackreed: For people who don't want to use GeoBlacklight-Schema, or if GeoBlacklight-Schema changes.

eliotjordan:
The presenter pattern in action! Taking business logic out of views seems much cleaner. This looks good. Just one question (see notes) and this PR is ready to sail.

lib/generators/geoblacklight/templates/catalog_controller.rb
eliotjordan: That makes sense. Especially for schema changes.

mejackreed:
:bow:

PR #352, opened 2015-11-04
Add detectRetina option for Leaflet basemaps and WMS layers

sgbalogh:
'detectRetina' documented [here](http://leafletjs.com/reference.html#tilelayer-detectretina):

> If true and user is on a retina display, it will request four tiles of half the specified size and a bigger zoom level in place of one to utilize the high resolution.


mejackreed:
Looks great, mind squashing 0539cdb into 3a8c9dc ?

sgbalogh:
Just overwrote my branch with a version that squashes all three commits into one. Not sure if I did it right... will I need to make another pull request?

mejackreed:
:+1:

PR #354, opened 2015-11-06
Update blacklight

mejackreed:
Allow more inclusive Blacklight versions, in preparation of Blacklight 6


eliotjordan:
:+1: Looks good.

PR #358, opened 2015-11-30
Adds simple IIIF generated JPEG download option

sgbalogh:
This pull request adds support for a "JPEG Download" option, nested under "Generated" in the download box, whenever a IIIF endpoint is available. It requests a full-sized JPEG image (/full/full/0/default.jpg) from that endpoint.

I really have no idea if I've implemented this in the best possible way, so please let me know what I can improve.

### DIFF
--- a/app/assets/javascripts/geoblacklight/modules/layer_opacity.js
+++ b/app/assets/javascripts/geoblacklight/modules/layer_opacity.js
@@ -4,11 +4,19 @@
   'use strict';
 
   L.Control.LayerOpacity = L.Control.extend({
-    initialize: function(layerGroup) {
-      var options = {
-        position: 'topleft',
-        layer: layerGroup.getLayers()[0]
-      };
+    initialize: function(layer) {
+      var options = { position: 'topleft' };
+
+      // check if layer is actually a layer group
+      if (typeof layer.getLayers !== 'undefined') {
+
+        // add first layer from layer group to options 
+        options.layer = layer.getLayers()[0];
+      } else {
+
+        // add layer to options
+        options.layer = layer;
+      }
 
       L.Util.setOptions(this, options);
     },
--- a/app/assets/javascripts/geoblacklight/viewers/map.js
+++ b/app/assets/javascripts/geoblacklight/viewers/map.js
@@ -47,6 +47,13 @@ GeoBlacklight.Viewer.Map = GeoBlacklight.Viewer.extend({
   },
 
   /**
+   * Add an opacity control to map.
+   */
+  addOpacityControl: function() {
+    this.map.addControl(new L.Control.LayerOpacity(this.overlay));
+  },
+
+  /**
   * Selects basemap if specified in data options, if not return mapquest
   */
   selectBasemap: function() {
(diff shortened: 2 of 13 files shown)


mejackreed:
@sgbalogh Thanks so much for this contribution! 

I think this approach will work, but we may need to shore up some of the tests. Also, this approach uses a cached version of the image downloaded locally to the GeoBlacklight server. Could we just use the direct file download link to the IIIF image, performance caching may not be an issue here, as it is with WMS, WFS.

@eliotjordan care to weigh in on the performance caching bit?

sgbalogh:
@mejackreed I'll try to make all of the changes as soon as I get a chance. Thanks for the helpful comments! Still very new to Rails, as you can probably tell...

About the direct download link option –– I considered doing exactly that, but I think at least for us at NYU there might be situations in which we would want to have the direct download link go to the preservation TIF master, stored in our digital repository, and so it would be nice to have another option for generating a compressed JPG copy. Though if we're alone in this use case then I guess there is no need to push this into the master.

eliotjordan:
@sgbalogh This is a great start! Thanks for the PR. In general, I like the shape of this. Definitely need some tests. As for the direct link vs. generated/cache, I think you have a good point. The 'original file' is probably a TIFF,  regardless of whether or not it is provided as a download. That said, I'm not sure if the derivative 'generated'  needs to be cached on the GBL server. IIIF servers generally use caching mechanisms already, and there might be permissions issues with doing this. I wonder if the download logic could be reconfigured to allow direct download of both.

Also, could you change  ```dc_format_s```  in the Princeton IIIF fixture to ```Scanned Map```? This will allow you to use it as a fixture for this functionality, and it's a more accurate description of the item.

sgbalogh:
@eliotjordan Let me look into what it would take to configure a direct download option, because you're right that it might be unnecessary. Do you think this would mean just abandoning use of the "download" class?

Also, would you mind pointing me to what you mean about the `dc_format_s

eliotjordan:
Oh sorry. I was referring to the [public_iiif_princeton.json](https://github.com/sgbalogh/geoblacklight/blob/0ee13368e14fe73cd9dae319afe4e5d654a7e1e0/spec/fixtures/solr_documents/public_iiif_princeton.json#L4) doc in spec/fixtures/. I'm not sure how to approach the the download functionality without taking a closer look. If you've got something in mind, give it a shot and let's see how it goes. I'll poke around a bit too.

sgbalogh:
Hi all –– I had another chance to work on this today, and look into a non-caching alternative. I agree that caching is not very useful, and potentially even really annoying for IIIF downloads... so the question becomes how to deal with directing the user to something like the '/full/full/0/default.jpg' image within the GeoBlacklight framework. I don't know the best way to do that if we're going to use the Download class for IIIF images... is there a good way for Ruby code in this class (or in the JpgDownload subclass) to initialize a browser redirect to an outside URL?

I guess another advantage to using the Download class is that if we ever wanted to build some sort of popup interface that allows users to configure the size and type of image they want, we could pass the request info into the Download as parameters.

eliotjordan:
@sgbalogh I just [pushed a branch](https://github.com/geoblacklight/geoblacklight/tree/iiif_download) on the geoblacklight repository with the commits from this PR rebased against master. I also added you as a collaborator on the project so you should have full write permissions. What do you think about closing out this PR and opening a new one from the branch I pushed?

sgbalogh:
@eliotjordan Sure, and thanks.

PR #362, opened 2015-12-23
remove Blacklight Range Limit as a dependency

mejackreed:
This PR removes the Blacklight Range Limit dependency. I think this should be ok by just requiring users who want to use this plugin to install it. Maybe a tutorial on Geoblacklight.org?


eliotjordan:
I good with this. Created an issue for the tutorial: #368.

PR #363, opened 2015-12-23
Moves around SolrDocument and DocumentPresenter to be more consistent with Blacklight

mejackreed:
This PR moves around SolrDocument and DocumentPresenter to be more consistent with Blacklight and Rails conventions


eliotjordan:
Good spruce up. A+ :100:

PR #371, opened 2016-01-08
Blacklight 6.x upgrade

mejackreed:
I'm not really happy with the way the routing turned out in 0577050d75e4100b120f0b8afbeb46cb8921223a and 0965390c01e4bd7cbbd301a912ea23a2d09069af. This probably needs some refinement.


eliotjordan:
@mejackreed In regards to your earlier comment about commits (with new hashes) a159de3 and 241b880, how would you improve these? And what is causing the issues with the engine not mounting routes?

mejackreed:
@eliotjordan Not sure.. was mostly focused on just getting it working. The routes would show up in `/rails/info` but wouldn't be accessible by the application? Maybe I wasn't referencing them in isolation correctly? ala `geoblacklight.downloads_path`?

I think the problem with merging as is, is the fact that we can't add routes down the road without an adopter adding them into their local application.

eliotjordan:
@mejackreed Yeah. That makes sense. I'm installing this branch into pulmap right now. I'll see what shakes out.

eliotjordan:
Spatial search isn't working for me on this branch. It looks as if the bbox param isn't being passed to solr.

eliotjordan:
Just need to add `include Geoblacklight::SpatialSearchBehavior` to the SearchBuilder model.

eliotjordan:
Uhh. Coverage decreased 33.7% after updating capybara and rspec-rails? Hmmm...

eliotjordan:
Princeton give this a 👍.

PR #373, opened 2016-01-14
Add rubocop

eliotjordan:
Adds rubocop to the project. Most of the fixes were minor. Below is a list of files/methods that had more complex changes.

- method name change
    - ``` lib/geoblacklight/wms_layer.rb:16 ```
    - ``` app/controllers/wms_controller.rb:3 ```
- remove already delegated function
    - ``` app/models/concerns/geoblacklight/solr_document.rb:30 ```
- refactor envelope to wsen
    - ``` app/models/concerns/geoblacklight/solr_document.rb:55 ```
- addresses the circular argument issue with params 
    - ``` lib/geoblacklight/view_helper_override.rb:14 :27 :30 ```
-  use let instead of instance variable
    - ``` spec/lib/geoblacklight/view_helper_override_spec.rb ```


spec/views/catalog/_document_split.html.erb_spec.rb
```diff
@@ -1,5 +1,5 @@
 require 'spec_helper'
 
-describe 'catalog/_document_split.html.erb' do
-  # TODO js plugin refactor really needs to be redone so that the views can be testable
+describe 'catalog/_document_split.html.erb' do # rubocop:disable RSpec/DescribeClass
```
mejackreed: Seems like we might run into this with any and all view specs. Could we add this rule in our `.rubocop.yml` file instead?

mejackreed:
A big :+1: overall. One question to ponder about including Rubocop in the default rake task. I would like to also do this (style checking) eventually with JavaScript and SCSS as well. I wonder if we could accomplish this using Hound rather than including a Rubocop task in the project?

I don't feel strongly either way here, I'm just curious about other approaches.

mejackreed:
I think I've convinced myself having Rubocop in is better than not :smile: ... What do you think about the rubocop yml level spec view rule?

eliotjordan:
Just moved the that rule into the config file. I've never used hound, but I have had success working on projects with Rubocop as a default rake task. It encourages good coding practices by forcing you to address issues on a regular basis.

PR #383, opened 2016-01-29
consolidate layerOptions to options, and use a global layer opacity d…

mejackreed:
…efault setting


eliotjordan:
ArcGIS is a go.

PR #384, opened 2016-01-30
Adds JPG download button when IIIF reference present

sgbalogh:
I took another quick stab at this. Not sure if it's the most elegant way of going about it... but it does bypass the download controller, so the "Download JPG" button will directly open a `/full/full/0/default.jpg` image in a new window.

Any thoughts? (Also, just to clarify, should I not have pushed this to a branch on the `geoblacklight/geoblacklight` repo? If so, I'll do this on a fork next time.)

### DIFF
--- a/app/views/catalog/_downloads.html.erb
+++ b/app/views/catalog/_downloads.html.erb
@@ -8,6 +8,8 @@
       <%= link_to(download_text(document.download_types.first[0]),
       download_hgl_path(id: document), data: {ajax_modal: 'trigger', download: 'trigger', download_type: 'harvard-hgl', download_id: document[:layer_slug_s] },
       class: 'btn btn-default') %>
+      <% elsif document.iiif_download.present? %>
+          <%= link_to "Download JPG", iiif_jpg_url, class: 'btn btn-default', download: 'trigger' %>
       <% else %>
       <%= link_to(download_text(document.download_types.first[0]), '', data: { download_path: "#{download_path(document[:layer_slug_s], type: document.download_types.first[0])}", download: 'trigger', download_type: document.download_types.first[0], download_id: document[:layer_slug_s] }, class: 'btn btn-default') %>
       <% end %>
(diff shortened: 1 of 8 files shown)


eliotjordan:
@sgbalogh Back in the office after a little vacay. I'll take a look at this tomorrow. As far as process goes, you are now a contributor, so you should absolutely push branches here. It makes things easier for others contributors to push commits to the PR.

sgbalogh:
@eliotjordan OK, good to know, thanks.

app/views/catalog/_downloads.html.erb
```diff
@@ -8,6 +8,8 @@
       <%= link_to(download_text(document.download_types.first[0]),
       download_hgl_path(id: document), data: {ajax_modal: 'trigger', download: 'trigger', download_type: 'harvard-hgl', download_id: document[:layer_slug_s] },
       class: 'btn btn-default') %>
+      <% elsif !@document.references.nil? && !@document.references.iiif.nil? %>
```
mejackreed: I wonder if this could be it's own partial?

Could we potentially move the logic to a helper method too?

sgbalogh:
@mejackreed added the logic checking to a geoblacklight_helper method that returns the jpg url from a IIIF endpoint. How does that look? For partials: do you mean having a partial just for the two links to the generated JPG, or for some larger part of the download interface?

mejackreed:
Maybe its not necessary now that things are moved to a helper. Could you add some tests for this functionality though? And maybe for the helper you added?

app/views/catalog/_downloads.html.erb
```diff
@@ -29,8 +31,16 @@
           <%= link_to(download_text(@document[:dc_format_s]), document.direct_download[:download], 'contentUrl' => document.direct_download[:download], data: { download: 'trigger', download_type: 'direct', download_id: document[:layer_slug_s] }) %>
         </li>
         <% end %>
-        <% if document.download_types.present? %>
+        <% if document.download_types.present? || document.iiif_download.present? %>
         <li role="presentation" class="dropdown-header">Generated</li>
+        <% end %>
+        <% if document.iiif_download.present? %>
```
mejackreed: I know the formatting in this file is already crazy.. mind indenting this block?
```erb
<% if document.iiif_download.present? %>
  <li>
    <%= link_to "Download JPG", iiif_jpg_url, download: 'trigger' %>
...
```

eliotjordan: Whoa! This is one gnarly partial. I think this could be refactored as Jack suggested, but perhaps that could happen in another pull request down the line.

eliotjordan:
@sgbalogh I think this will be a useful feature. If you need any test writing advice, give us a shout.  Looks good and plays well with records that have download urls:

screenshot 2016-02-10 12 49 20

eliotjordan:
@sgbalogh :clap: Nice job on the specs. I reworked the download helper spec a bit and pushed it up in a separate commit. Feel free to do what you want with it.

eliotjordan:
About the reworked spec. The main change here is to use **assign** to set @document. Also, this commit makes the solr document definition a bit more readable and changes "/images/" in the IIIF url to "/image/" to reflect that you are referencing a single image with this endpoint.

mejackreed:
Looks good! Could we can squash down all of these commits?

sgbalogh:
Just squashed them all into one commit–– didn't realize this would also remove any record of the contribution by @eliotjordan though...  what is the standard practice here?

Also, thanks to both of you for all of the assistance with this –– this has definitely taught me a lot about how to properly contribute to the application.

mejackreed:
:clap:

aeschylus:
Very cool.

Issue #386, opened 2016-02-22
CartoDB OneClick failure with multi-word "provider" attribute


sgbalogh:
Recently changed the application name of our deployment to "NYU Spatial Data Repository", which breaks the "Open in CartoDB" feature; as it turns out, the provider attribute in the OneClick URL does not like multi-word values, regardless of whether the spaces are URL encoded with `+` or `%20`. I played with a few different variations, and can only get CartoDB to recognize the input file when the provider is a single word (or altogether absent).

[Here](https://github.com/geoblacklight/geoblacklight/blob/c4a1d934e482aad2101f83d1dc5038d6c516875c/app/helpers/geoblacklight_helper.rb#L115) is the line responsible for the OneClick param in question. 

Has anyone run into anything like this before? Easiest solution might be to remove all blank space from `application_name` before sending. Could we also just omit this param altogether? Not sure what CartoDB's functionality for `provider` actually is... and relatedly, if it is meant for layer citation purposes, then maybe we should actually be sending the institution as the provider, rather than the application name?

mejackreed:
Also may be worth filing a bug report with CartoDB. They have been responsive on other bug reports.

PR #387, opened 2016-02-22
Fixes CartoDB OneClick multi-value application name bug

sgbalogh:
Removes any whitespace from `application_name` before sending it as the `provider` param.

Referenced in https://github.com/geoblacklight/geoblacklight/issues/386

Issue on the CartoDB side already documented here https://github.com/CartoDB/cartodb/issues/6092

### DIFF
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -106,13 +106,20 @@ module GeoblacklightHelper
   end
 
   ##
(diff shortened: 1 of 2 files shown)


mejackreed:
:cool: It might be good to add a test here to make sure that this is happening (so someone doesn't remove it).

Potentially could also be moved to an additional method?

sgbalogh:
@mejackreed, how does this new commit look?

mejackreed:
Looks good!

PR #388, opened 2016-02-22
Leaflet full screen control

eliotjordan:
Adds full screen control to Leaflet viewers. It actually does pseudo full screen, but that can be changed to real full screen. I'm adding this to the IIIF viewer in our local Geoblacklight and thought it might be useful to others. Thoughts?

If not in the vanilla geoblacklight, perhaps this could be a plugin gem. Or at least a blog write-up if others want to implement the feature.


sgbalogh:
@eliotjordan, this is great! Just had a chance to try it out. I definitely think it would be a useful addition, particularly for IIIF images.

mejackreed:
Looks good! I'm wondering if there is a way that we could make this configurable. Though I like the plugin approach, it also feels like that might be a lot of setup to get going. Configurability might be a simpler solution?

eliotjordan:
@mejackreed Making this a configuration option is a good idea. I'll see if I can whip something up.

eliotjordan:
I finally had some time to circle back to this. A few changes:
*   added a section in the setting.yml file for leaflet plugin settings
*   these settings are passed to viewer javascript files using a data attribute
*   added an addControls to viewer.js to add leafet controls all at once
*   full screen and opacity controls now use leaflet plugin settings to check if they should be displayed on a particular viewer

app/assets/javascripts/geoblacklight/viewers/iiif.js
```diff
@@ -21,5 +21,5 @@ GeoBlacklight.Viewer.Iiif = GeoBlacklight.Viewer.extend({
 
     // expand viewer element
     $(this.element).parent().attr('class', 'col-md-12');
-  }
+  },
```
eliotjordan: I was using an airbnb style linter, and kinda waffled back and forth on these. I can remove them for the sake of compatibility, but as of January, [IE8 is no longer supported by Microsoft](https://support.microsoft.com/en-us/lifecycle#gp/msl-ie-dotnet-an). Have any browser metrics for Earthworks or NYU GeoBL?

mejackreed:
Regarding trailing commas. I am +1 for removing them. It is definitely part of the Airbnb Javascript ES6 styleguide, but not part of the ES5 one.

## ES6
https://github.com/airbnb/javascript#commas

## ES5
https://github.com/airbnb/javascript/tree/master/es5#commas

I would propose until GeoBlacklight updates to ES6, lets stick with the ES5 styleguide. I'll work on adding a PR that adds a `.jshintrc` for AirBnb's ES5 styleguide... :cool:?

eliotjordan:
Removed those pesky trailing commas.

eliotjordan:
Superseded by #400. See issue #402.

Issue #393, opened 2016-03-24
Simplified Metadata Schema for v1.0


drh-stanford:
See https://github.com/geoblacklight/geoblacklight-schema for current schema. See #423 for v2.0 changes.

Tasks:
- [x] Update documentation for GeoBlacklight-Schema -- https://github.com/geoblacklight/geoblacklight-schema/pull/89

Deprecate:
- [x] Deprecate `uuid` -- see https://github.com/geoblacklight/geoblacklight-schema/issues/53
- [x] Deprecate `dc_relation_sm` (keep `dct_isPartOf_sm ` which is currently being used for collections) -- see https://github.com/geoblacklight/geoblacklight-schema/issues/85
- [x] Deprecate `solr_year_i` (only used by the Blacklight range plugin, not core GeoBlacklight, and generally you want a multi-valued field here) -- see https://github.com/geoblacklight/geoblacklight-schema/issues/90
- [x] Deprecate `georss_box_s`, `georss_point_s` and `georss_polygon_s` (all unused)

Adding:
- [x] Add `dc_source_sm` for derived layers -- see #412 and https://github.com/geoblacklight/geoblacklight-schema/issues/83
- [x] Add `geoblacklight_version ` for schema version of record

mejackreed:
Maybe now `*_srpt` ?

mejackreed:
Also make sure to loop in #412 and https://github.com/geoblacklight/geoblacklight-schema/issues/83

drh-stanford:
possibly drop `dc_relation_sm` and keep `dct_isPartOf_sm`

drh-stanford:
For changing/relocating, we can 
* [ ] submit PR to OpenGeoMetadata
* [ ] change GeoCombine
* [ ] add directions in GeoBlacklight upgrade guide
  * [ ] each institution would implement

andrewbattista:
This looks good! I think I was talking with @sgbalogh yesterday about some of the implications of this

drh-stanford:
To deprecate `uuid` and `dc_identifier_s` in favor of `layer_slug_s` we need to update the Solr `schema.xml` here:

https://github.com/geoblacklight/geoblacklight/blob/master/schema/solr/conf/schema.xml#L3

and switch `uuid` from required = `true` to `false`

drh-stanford:
[...]

PR #395, opened 2016-03-24
Make GeoBlacklight::Presenter helper method more generalizable with m…

mejackreed:
…ethod message pass

Part of the metadata schema abstraction work. Prerequisite for #393


mejackreed:
The Hound comments should be addressed in #396

tampakis:
Hound is happy now.

PR #399, opened 2016-03-24
More field abstraction

tampakis:


eliotjordan:
This looks good to me.

PR #400, opened 2016-03-25
Add support for adding controls from within a GeoBlacklight application

mejackreed:


mejackreed:
I've also tested this locally in adding the fullscreen plugin. Works great!

Should we add a wiki guide or something on the website?

app/assets/javascripts/geoblacklight/viewers/viewer.js
```diff
@@ -8,7 +12,44 @@ GeoBlacklight.Viewer = L.Class.extend({
...
+
+    /**
+    * Loop though the GeoBlacklight.Controls hash, and for each control,
+    * check to see if it is included in the controls list for the current
+    * viewer. If it is, then pass in the viewer object and run the function
+    * that adds it to the map.
+    **/
+    $.each(GeoBlacklight.Controls, function(name, func) {
+      if (controls && controls.indexOf(name) > -1) {
+        func(_this);
```
gravesm: It would probably be more javascript-y to explicitly set the context on the function call:

`func.call(_this)`

which would mean the viewer becomes `this` in a control function.

mejackreed: Probably.. I like that approach.

mejackreed: Hey @gravesm updated this pr to use the `func.call(_this)` approach. Also modified in `opacity.js`

eliotjordan:
This looks good. @sgbalogh and/or @drh-stanford, wanna take a final look? Jack, Mike, and I have had our grubby fingers all over this.

sgbalogh:
Looks good to me! Should be a very useful addition.

eliotjordan:
Is this puppy ready to :roller_coaster: @sgbalogh?

PR #403, opened 2016-04-07
uses simplified MIME type of download for comparison

sgbalogh:
To get around an error in which GeoServer returns a download with a Content-Type that is correct, but more detailed than what was requested. Like when `application/json` is requested but the download from GeoServer is `application/json; chartype=UTF-8`

### DIFF
--- a/lib/geoblacklight/download.rb
+++ b/lib/geoblacklight/download.rb
@@ -40,11 +40,8 @@ module Geoblacklight
     def create_download_file
       download = initiate_download
       File.open("#{file_path_and_name}.tmp", 'wb') do |file|
-        if download.headers['content-type'] == @options[:content_type]
-          file.write download.body
-        else
-          fail Geoblacklight::Exceptions::WrongDownloadFormat
-        end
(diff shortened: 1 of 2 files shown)


lib/geoblacklight/download.rb
```diff
@@ -33,14 +33,18 @@ def get
       end
     end
 
+    def reveal_mimetype(download)
```
mejackreed: Could this be a private method? And I wonder if you could put all of the logic in a method like `matches_mimetype`?

```ruby
if matches_mimetype(download)
  file.write download.body
...

private
def matches_mimetype(download)
  MIME::Type.simplified(download.headers['content-type']) == @options[:content_type]
end
```

You may not be able to unit test this method, but you could add a case to the `#create_download` block?

Just thinking here about limiting our exposure of the public api. Would someone need to modify this method?

mejackreed:
Minor comment... Not sure if we need to have it public? Otherwise looks awesome! :+1:

mejackreed:
:fireworks:

PR #407, opened 2016-07-05
make Cartodb the default postiron basemap

mejackreed:
resolves the issue where the MapQuest OSM basemap is going away.


mejackreed:
Hmm.. doesn't seem like travis ran here.

eliotjordan:
Looks good. RIP Mapquest tiles.

PR #416, opened 2016-07-25
Solr6

### LINKED ISSUE #409: Update to work with Solr 5 & 6
- [ ] update to Solrwrapper
 - [ ] update rake tasks (https://github.com/geoblacklight/geoblacklight/blob/master/lib/tasks/configure_solr.rake#L8)
 - [ ] update document + tutorials
 - [ ] update `solrconfig.xml` copier to insert the correct `lucenceMatchVersion`

cbeer:
Fixes #409


eliotjordan:
Finally ready to go on this. When the PR is merged, I'll update the wiki.

eliotjordan:
To run solr and rails server, use this command:
```
bundle exec rake geoblacklight:server
```

No more switching back and forth between to the test app to load fixtures etc. This command also works in an application generated from the template.

solr/conf/schema.xml
```diff
@@ -143,10 +143,8 @@
     <fieldType name="location" class="solr.LatLonType" subFieldSuffix="_d"/>
 
     <fieldType name="location_rpt" class="solr.SpatialRecursivePrefixTreeFieldType"
-               distErrPct="0.025"
-               maxDistErr="0.000009"
-               units="degrees"
-            />
+               geo="true" distErrPct="0.025" maxDistErr="0.001" distanceUnits="kilometers"/>
+
```
drh-stanford: do we use distance filtering in GeoBlacklight? the change in units here might be problematic if we do.

mejackreed: No we don't.. but potentially this is a breaking change if someone is and they use this schema. This is totally configurable per installation.

eliotjordan: We can revert to degrees if we explicitly set `geo="false"`.

solr/conf/schema.xml
```diff
@@ -143,10 +143,8 @@
     <fieldType name="location" class="solr.LatLonType" subFieldSuffix="_d"/>
 
     <fieldType name="location_rpt" class="solr.SpatialRecursivePrefixTreeFieldType"
-               distErrPct="0.025"
-               maxDistErr="0.000009"
-               units="degrees"
-            />
+               geo="true" distErrPct="0.025" maxDistErr="0.001" distanceUnits="degrees"/>
```
eliotjordan: Fixed.

solr/conf/schema.xml
mejackreed: Why revert back? If the schema is versioned alongside the code.. I'm ok with it. `degrees` is more relevant to non-geo spatial distances, I think `kilometers` is more consistent with geo ones? Discuss on slack?

eliotjordan:
@mejackreed You're correct about using kilometers instead of degrees. I didn't read the `SpatialRecursivePrefixTreeFieldType` documentation as closely as I should. The warren of Confluence pages makes my head hurt:

> The most up to date information about spatial search at the Solr Reference Guide instead: https://cwiki.apache.org/confluence/display/solr/Spatial+Search. Some details, examples, and other info is still here, however.

I wonder why the units were set to degrees in the earlier version.

solr/conf/solrconfig.xml
```diff
@@ -20,11 +20,25 @@
...
-  <luceneMatchVersion>4.7</luceneMatchVersion>
-  <dataDir>${solr.data.dir:}</dataDir>
+  
+  <!-- Controls what version of Lucene various components of Solr
+       adhere to.  Generally, you want to use the latest version to
+       get all bug fixes and improvements. It is highly recommended
+       that you fully re-index after changing this setting as it can
+       affect both how text is indexed and queried.
+  -->
+  <luceneMatchVersion>5.0.0</luceneMatchVersion>
```
mejackreed: Is this the correct lucene  version?

mejackreed:
Great work here overall. I think this might constitute a breaking change in GeoBlacklight though. I wonder is it even possible to implement this in the way that Blacklight did, with allowing for the `Jettywrapper` option?

solr/conf/solrconfig.xml
eliotjordan: As far as I know, there isn't any Solr6 specific configuration here. This will work on version 5 as well as 6.

eliotjordan:
What if we include a solr4 version of the schema for backwards compatibility with existing systems? Again, I don't think we use any solr5 or solr6 specific functionality. As far as including `Jettywrapper`, I'm not sure I see the utility. No one should be using it in production, and the `solr wrapper`-based workflow is much more simple.

solr/conf/solrconfig.xml
drh-stanford: Based on our discussion earlier, I'd prefer to see this set to 6.1.0.

mejackreed:
When trying to use this locally in EarthWorks I get the error:

```sh
NameError: uninitialized constant SolrWrapper
~/dev/geoblacklight/lib/tasks/geoblacklight.rake:7:in `block (2 levels) in <top (required)>'
~/.rbenv/versions/2.2.3/bin/bundle:23:in `load'
~/.rbenv/versions/2.2.3/bin/bundle:23:in `<main>'
Tasks: TOP => geoblacklight:server
(See full trace by running task with --trace)
```

lib/tasks/geoblacklight.rake
```diff
@@ -2,7 +2,16 @@ require 'rails/generators'
...
+  desc 'Run Solr and GeoBlacklight for interactive development'
+  task :server, [:rails_server_args] do |_t, args|
+    SolrWrapper.wrap(port: '8983') do |solr|
+      solr.with_collection(name: 'blacklight-core', dir: File.join(File.expand_path('../../', File.dirname(__FILE__)), 'solr', 'conf')) do
+        system "bundle exec rails s #{args[:rails_server_args]}"
+      end
+    end
+  end
+
+  namespace :index do
```
mejackreed: Can we provide an alias here back to `geoblacklight:solr:seed`? That task is missing and potentially downstream clients may be using it.

eliotjordan:
@mejackreed Re-added `geoblacklight:solr:seed` task. Any other tasks to restore?

PR #426, opened 2016-07-26
migrate v0.3.0 GeoBlacklight Schema into local repo

### LINKED ISSUE #425: Move geoblacklight required files from geoblacklight-schema
The `geoblacklight-schema` repo is historical and has tools/examples that are not used by GeoBlacklight. This issue is to migrate those files that GeoBlacklight *does* use into the geoblacklight repo. Be *sure* to use Solr 4 files -- see https://github.com/geoblacklight/geoblacklight/issues/409 for Solr 6 issues

- [ ] `schema.xml`
- [ ] `solrconfig.xml`
- [ ] schema doc

Note that there are 2 pathways:
- configuring a development instance of Solr for GeoBlacklight
- configuring a production instance of Solr for GeoBlacklight

drh-stanford:
This PR fixes #425. It creates a `schema/` folder in the repo and puts the Solr configuration files and the documentation on the schema there. 

The rake task `geoblacklight:configure_solr` now takes an optional argument for the Solr version.

### DIFF
--- a/lib/tasks/configure_solr.rake
+++ b/lib/tasks/configure_solr.rake
@@ -1,36 +1,27 @@
 require 'rails/generators'
 require 'generators/geoblacklight/install_generator'
 require 'geoblacklight'
-require 'open-uri'
+require 'fileutils'
 
 namespace :geoblacklight do
-  desc 'Copies the default SOLR config for the included Solr'
+  desc 'Configures Solr for local jetty instance'
   task :configure_solr do
-    files_urls = [
+    root = Gem::Specification.find_by_name('geoblacklight').gem_dir rescue '.'
+    [
       {
-        url: 'https://raw.githubusercontent.com/geoblacklight/geoblacklight-schema/v0.3.0/conf/schema.xml',
+        src: "schema/solr/conf/schema.xml",
         file: 'schema.xml'
       },
       {
-        url: 'https://raw.githubusercontent.com/geoblacklight/geoblacklight-schema/v0.3.0/conf/solrconfig.xml',
+        src: "schema/solr/conf/solrconfig.xml",
         file: 'solrconfig.xml'
       }
-    ]
-    files_urls.each do |item|
-      puts item.inspect
-      begin
-        open(item[:url]) do |io|
-          IO.copy_stream(io, "jetty/solr/blacklight-core/conf/#{item[:file]}")
-        end
-      rescue Exception => e
-        abort "Unable to download #{item[:file]} from #{item[:url]} #{e.message}"
-      end
+    ].each do |item|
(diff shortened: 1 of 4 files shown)


lib/tasks/configure_solr.rake
```diff
@@ -1,36 +1,27 @@
 require 'rails/generators'
 require 'generators/geoblacklight/install_generator'
 require 'geoblacklight'
```
drh-stanford: i'm not sure whether these requires ^^^ are necessary...

lib/tasks/configure_solr.rake
```diff
@@ -1,36 +1,27 @@
...
-  desc 'Copies the default SOLR config for the included Solr'
-  task :configure_solr do
-    files_urls = [
+  desc 'Configures Solr for local jetty instance'
+  task :configure_solr, [:version] do |_t, args|
+    args.with_defaults(version: '4.7')
+    [
       {
-        url: 'https://raw.githubusercontent.com/geoblacklight/geoblacklight-schema/v0.3.0/conf/schema.xml',
+        src: "schema/solr/#{args[:version]}/conf/schema.xml",
```
mejackreed: Have we tried this on a GeoBlacklight app, not the engine_cart one?

drh-stanford: No, I only tried it from engine_cart, locally and via Travis. Does it need to go in the generator?

mejackreed: Not sure.. I'll take it for a spin locally and see if it works.

mejackreed: Looks like it does break. Maybe something like:

`Gem::Specification.find_by_name("your_gem_name").gem_dir`

or 

http://stackoverflow.com/a/10133837

I think you may have `Blacklight.root` available to you?

mejackreed:
Curious about the version directory here. Could we be ok with not having a directory and maintaining versions of the solr schema for each version?

drh-stanford:
This version in the directory path mimics the `luceneMatchVersion` here -- https://github.com/geoblacklight/geoblacklight/blob/b4568633db752ddcea08101b1735acc8976a6b0f/schema/solr/4.7/conf/solrconfig.xml#L23

If we want to support several versions of Solr, we need some way to write out the correct version number there -- and possibly other changes to the `solrconfig.xml`. I can remove the version in the path and introduce some mechanism later when we do #409.

mejackreed:
Gotcha.. yeah I understand that. I'm concerned that this sets the precedent that we will maintain these schemas for any version of Solr when in reality they don't change much except for `luceneMatchVersion`. /shrug I'm ok with either direction.

drh-stanford:
I think I've addressed all the comments. Marked WIP until we can test it on a generated geoblacklight app.

mejackreed:
It works!

PR #433, opened 2016-07-27
CartoDB basemaps display higher resolution on retina screen

eliotjordan:
This PR makes basemaps higher resolution when displayed on a retina monitor. Not sure how I feel about it. They look very sharp (obvs), but are harder to read and take a noticeably longer amount of time to load.

### DIFF
--- a/app/assets/javascripts/geoblacklight/basemaps.js
+++ b/app/assets/javascripts/geoblacklight/basemaps.js
@@ -2,19 +2,21 @@
 
 GeoBlacklight.Basemaps = {
   darkMatter: L.tileLayer(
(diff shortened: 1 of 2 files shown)


eliotjordan:
...and every feature test is broken.

mejackreed:
hmm.. I say let an adopter do this in override type way

mejackreed:
Also could this work instead? https://github.com/leaflet-extras/leaflet-providers#retina-tiles

Using `{retina}`

eliotjordan:
Thanks for the tip @mejackreed. Force pushed a new version of this PR that is far simpler. Adopters can simply indicate that they want to use retina basemaps by overriding `GeoBlacklight.Basemaps` and setting `detectRetina: true`.

PR #445, opened 2016-07-27
update fake esri fixtures with real provenance

eliotjordan:
Update fake esri fixtures with real provenance so that icons show up properly.


eliotjordan:
I assign the ImageMapLayer fixture to Princeton, because we will likely have real image services in geoblacklight in the future. And NYU looked lonely without a fixture of it's own.

PR #447, opened 2016-07-28
fixes our facet icon alignment issue

### LINKED ISSUE #438: Facet display with geoblacklight-icon issue
See below for example:
screen shot 2016-07-26 at 3 21 00 pm

This seems to have to do with [this code](https://github.com/projectblacklight/blacklight/blob/master/app/assets/stylesheets/blacklight/_facets.scss#L77-L84), part of Blacklight core.

Changing the above to:
```
  .facet-label {
    display: inline-block;
    padding-right: 1em;
    margin-left: 5px;
    text-indent: -15px;
    padding-left: 15px;
    padding-bottom: $padding-base-vertical;
    @include hyphens-auto;
  }
```
seems to fix it adequately. See below:

screen shot 2016-07-27 at 11 39 08 am

mejackreed:
Not really sure how to test this.. but fixes #438 

@sgbalogh what do you think? The "Restricted" / "Public" icons still seem off?

screen shot 2016-07-27 at 5 07 30 pm

### DIFF
--- a/app/assets/stylesheets/geoblacklight/_geoblacklight.scss
+++ b/app/assets/stylesheets/geoblacklight/_geoblacklight.scss
@@ -15,6 +15,7 @@
 @import 'font-awesome';
 @import 'geoblacklight/styles';
 @import 'modules/coderay';
+@import 'modules/facets';
 @import 'modules/mixins';
 @import 'modules/icon-customization';
 @import 'modules/home';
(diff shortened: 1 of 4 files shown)


sgbalogh:
Nice! I'm not that bothered by the padlocks being off... my fix actually has a similar problem anyway:

screen shot 2016-07-28 at 11 31 24 am

eliotjordan:
👍

PR #455, opened 2016-07-29
remove references to stopwords synonyms and protwords.txt files

### LINKED ISSUE #453: Solr conf is missing some files
The new schema.xml is looking for `protwords.txt`, `stopwords_en.txt` and `synonyms.txt`, which don't exist.

drh-stanford:
This PR fixes #453 . These filters are not really used.


mejackreed:
I think `stopwords` is probably used by some. Could we just add these files?

eliotjordan:
I second @mejackreed. Let's add the files.

drh-stanford:
Ok, I've pushed the files to the branch and left `schema.xml` unchanged.

PR #456, opened 2016-07-29
Rails 5 Compatibility 

eliotjordan:
Initial experiments with rails 5 compatibility. There are 11 broken tests and a slew of deprecation warnings, but the app seems to run just fine. Here are the steps I used to get this running:

1. Start with a clean gemset and switch to the `rails-5` branch.
1. Run `RAILS_VERSION=5.0.0 bundle install`
1. Monkey patch rails to fix error with the squish method.
    - Not sure what this is about, but this can be tracked down later.
    - This is the [line in actionpack](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/routing.rb#L257) that needs to be patched.
   - Find that file in your gemset. For me, this was the path:  
     
     ```
     /Users/eliotj/.rvm/gems/ruby-2.2.2@r5-bl/gems/actionpack-5.0.0/lib/action_dispatch/routing.rb`
     ```
  - Using the `gem which rails` command can help you find the gem location.
  - Remove the `.squish` method so that line 257 looks like:

    ```
    INSECURE_URL_PARAMETERS_MESSAGE = <<-MSG
    ```
1. Then execute the standard set of commands to get GeoBlacklight running.

```
bundle exec rake engine_cart:generate
bundle exec rake jetty:clean
bundle exec rake geoblacklight:configure_solr
bundle exec rake jetty:start
cd .internal_test_app/
rake geoblacklight:solr:seed
rails s
```

### DIFF
--- a/geoblacklight.gemspec
+++ b/geoblacklight.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'blacklight', '~> 6.3'
   spec.add_dependency 'leaflet-rails', '~> 0.7.3'
   spec.add_dependency 'font-awesome-rails'
-  spec.add_dependency 'config', '~> 1.2.1'
+  spec.add_dependency 'config'
   spec.add_dependency 'faraday'
   spec.add_dependency 'coderay'
   spec.add_dependency 'geoblacklight-icons', '>= 0.2'
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec-rails', '~> 3.1'
   spec.add_development_dependency 'engine_cart', '~> 0.10'
   spec.add_development_dependency 'capybara', '>= 2.5.0'
-  spec.add_development_dependency 'poltergeist', '>= 1.5.0'
+  spec.add_development_dependency 'poltergeist'
   spec.add_development_dependency 'factory_girl_rails'
   spec.add_development_dependency 'database_cleaner', '~> 1.3'
 end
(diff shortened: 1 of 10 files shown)


mejackreed:
Err.. yeah.. no likey the modify gemset :)

I was talking to @cbeer  about this, I can't remember what they are doing for Spotlight.. but https://github.com/projectblacklight/spotlight/pull/1483

eliotjordan:
No longer an experimental branch. After #416 is merged, this branch must be rebased against it. At that point, all tests should pass.

eliotjordan:
Made some updates to improve backwards compatibility with Rails 4. Thanks to @cbeer for the `BackportTestHelpers` module. That saved tons of time and got rid of those nasty ActionController::TestCase deprecation warnings for good.

geoblacklight.gemspec
```diff
@@ -18,22 +18,22 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
+  spec.add_dependency 'rails', '>= 4.2.0', '< 6'
   spec.add_dependency 'blacklight', '~> 6.3'
   spec.add_dependency 'leaflet-rails', '~> 0.7.3'
   spec.add_dependency 'font-awesome-rails'
-  spec.add_dependency 'rails_config', '~> 0.4.2'
+  spec.add_dependency 'config', '~> 1.2.1'
```
cbeer: This dependency pinning strikes me as overly pessimistic.

Issue #424, opened 2016-07-26
Direct download for documentation/codebook from dct:references


sgbalogh:
Any interest in something like this for core? At NYU we are using `http://schema.org/additionalType` as a key, in `dct_references_s`, with a value that is a URL to a direct download link for some sort of documentation related to a record. An example of a record with documentaiton is here: https://geo.nyu.edu/catalog/nyu_2451_34670

This is also reliant on being able to specify URI constants.

mejackreed:
Closed by #461 ?

PR #461, opened 2016-08-01
Adds URI constant for data_dictionary (/documentation download)

sgbalogh:
https://github.com/geoblacklight/geoblacklight/issues/424

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
@@ -22,6 +22,10 @@
   @extend .fa, .fa-bookmark;
 }
 
+.geoblacklight-data-dictionary {
+  @extend .fa, .fa-book;
+}
+
 .geoblacklight-sms {
   @extend .fa, .fa-mobile;
 }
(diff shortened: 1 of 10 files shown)


mejackreed:
Awesome! Is there any UI functionality we should add also in this PR?

Potentially also update http://geoblacklight.org/tutorial/2015/02/09/geoblacklight-overview.html?

sgbalogh:
Happy to add a tool partial for it, if everyone is OK with that being part of GeoBlacklight core: 
screen shot 2016-08-01 at 4 35 34 pm

Any suggestions for default text? Perhaps just `Data Dictionary`, or even more generally just `Documentation`?

mejackreed:
👍 all the way on your suggestions.

drh-stanford:
maybe "Detailed Documentation" or just "Documention"

sgbalogh:
@drh-stanford in the PR the [default](https://github.com/geoblacklight/geoblacklight/blob/376e99566269c1cfd74c0e19142185ca1d8ad02f/config/locales/geoblacklight.en.yml#L33) is set to `Documentation`; is that ok?

eliotjordan:
Excellent work. ✨

PR #463, opened 2016-08-01
Splits out download button partials for better understanding

mejackreed:


eliotjordan:
Nice refactor. Much easier to follow.

Issue #471, opened 2016-08-03
Change CartoDB to Carto


andrewbattista:
As per Carto's re-branding, can we change the button to say "open in Carto" instead of "open in CartoDB?"

eliotjordan:
Closed by #472.

PR #472, opened 2016-08-03
Moves carto text to locale, changes to 'Carto'

sgbalogh:
Addresses https://github.com/geoblacklight/geoblacklight/issues/471

Did this in as minimal of a way possible; didn't want to touch any of the methods with `cartodb` in the name, etc. Perhaps that is something we can do for a 2.0.


lib/generators/geoblacklight/templates/settings.yml
```diff
@@ -1,8 +1,8 @@
 # Configurable Logo Used for CartoDB export
 APPLICATION_LOGO_URL: 'http://geoblacklight.org/images/geoblacklight-logo.png'
 
-# CartoDB OneClick Service https://cartodb.com/open-in-cartodb/
-CARTODB_ONECLICK_LINK: 'http://oneclick.cartodb.com/'
+# Carto OneClick Service https://carto.com/engine/open-in-carto/
+CARTO_ONECLICK_LINK: 'http://oneclick.cartodb.com/'
```
eliotjordan: Is the url still `http://oneclick.cartodb.com/`? Looks `http://oneclick.carto.com/` might work as well.

mejackreed: Hmm... maybe? I was going off of the documentation at https://carto.com/engine/open-in-carto/

screen shot 2016-08-05 at 11 34 05 am

eliotjordan: Just tried it. Works with `carto.com` as well. But I'm OK leaving it as is because that's what's in the docs.

PR #477, opened 2016-08-04
Improve strictness of schema validation

drh-stanford:
This PR improves the enum and pattern matching validation. It also makes `dc_format_s` a free-text field to support more file formats. This PR is connected to https://github.com/geoblacklight/geoblacklight-schema/issues/84


eliotjordan:
This looks good to me.  👍

PR #482, opened 2016-08-05
Add the gem release task back to the Rakefile

eliotjordan:
Adds the back the gem release rake task.


mejackreed:
why no? `require 'bundler/gem_tasks'`

eliotjordan:
That works too. I guess `require 'rubygems'` is redundant? This pattern came from Blacklight. Either way is fine with me.

PR #487, opened 2016-08-10
Fix WMS feature inspection

eliotjordan:
Fix WMS feature inspection. ~~Rails 5 parameters are objects and need to be converted to hashes using the `to_unsafe_h` method for compatibility.~~ We have to explicitly permit WMS parameters (url, layer, bbox, etc…).

### DIFF
--- a/spec/lib/geoblacklight/wms_layer_spec.rb
+++ b/spec/lib/geoblacklight/wms_layer_spec.rb
@@ -1,28 +1,63 @@
 require 'spec_helper'
 
 describe Geoblacklight::WmsLayer do
-  let(:params) { { 'URL' => 'http://www.example.com/', 'X' => '277' } }
   let(:wms_layer) { described_class.new(params) }
+  let(:rails_4_params) { { 'URL' => 'http://www.example.com/', 'X' => '277' } }
+  let(:rails_5_params) { instance_double('ActionController::Parameters') }
+
+  before do
+    allow(rails_5_params).to receive(:to_h).and_return('URL' => 'http://www.example.com/', 'X' => '277')
+  end
+
   describe '#initialize' do
+    let(:params) { rails_4_params }
+
     it 'initializes as a WmsLayer object' do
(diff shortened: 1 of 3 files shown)


spec/lib/geoblacklight/wms_layer_spec.rb
```diff
@@ -1,28 +1,63 @@
...
 
 describe Geoblacklight::WmsLayer do
-  let(:params) { { 'URL' => 'http://www.example.com/', 'X' => '277' } }
   let(:wms_layer) { described_class.new(params) }
+  let(:rails_4_params) { { 'URL' => 'http://www.example.com/', 'X' => '277' } }
+  let(:rails_5_params) { double('ActionController::Parameters') }
+
+  before do
+    allow(rails_5_params).to receive(:to_unsafe_h).and_return('URL' => 'http://www.example.com/', 'X' => '277')
+  end
```
drh-stanford: the last bit can be `and_return(rails_4_params)`

lib/geoblacklight/wms_layer.rb
```diff
@@ -1,6 +1,7 @@
 module Geoblacklight
   class WmsLayer
     def initialize(params)
+      params = params.to_unsafe_h unless params.is_a?(Hash)
       @params = params.to_h.merge(Settings.WMS_PARAMS)
```
drh-stanford: doesn't the `params.to_h` already convert the params object into a Hash? maybe add a comment as to why `to_unsafe_h` is needed here

spec/lib/geoblacklight/wms_layer_spec.rb
eliotjordan: 👍

lib/geoblacklight/wms_layer.rb
eliotjordan: Good point. We should explicitly permit the params. I'll move this into the WMS controller.

PR #488, opened 2016-08-10
Pin the geoblacklight gem version in template.rb

eliotjordan:
Pin the geoblacklight gem version in the template file. There have been some issues with users running the `rails new` command with older versions of the gem installed.

### DIFF
--- a/template.rb
+++ b/template.rb
@@ -1,5 +1,5 @@
-gem 'blacklight'
-gem 'geoblacklight'
+gem 'blacklight', '>= 6.3'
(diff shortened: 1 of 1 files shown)


template.rb
```diff
@@ -1,5 +1,5 @@
-gem 'blacklight'
-gem 'geoblacklight'
+gem 'blacklight', '>= 6.4'
```
mejackreed: Should this match the gemspec?

eliotjordan: Yes it should.

PR #497, opened 2016-08-24
Remove non alpha numeric

mejackreed:
This will enable more adopters, potentially with alpha numeric chars in their names. eg "Lewis & Clark"

### DIFF
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -39,7 +39,7 @@ module GeoblacklightHelper
   def geoblacklight_icon(name)
     content_tag :span,
                 '',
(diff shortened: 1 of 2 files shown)


app/helpers/geoblacklight_helper.rb
```diff
@@ -39,7 +39,7 @@ def render_facet_tags(facet)
   def geoblacklight_icon(name)
     content_tag :span,
                 '',
-                class: "geoblacklight-icon geoblacklight-#{name.downcase.tr(' ', '-')}",
+                class: "geoblacklight-icon geoblacklight-#{name.parameterize.underscore.dasherize}",
```
cbeer: Is this equivalent to `name.parameterize(separator: '-')`?

mejackreed: Yeah seems that way. Thanks for catching that.

PR #498, opened 2016-09-01
Support both a single layer and an array of layers

### LINKED ISSUE #496: JSON Schema has wrong type
I may be missing something, but it looks like the schema has the wrong type associated with it, on line 6:

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "description": "Schema for GeoBlacklight. See https://github.com/geoblacklight/geoblacklight/wiki/Schema for more details.",
  "id": "http://geoblacklight.org/v1.0/schema",
  "title": "GeoBlacklight Schema",
  "type": "array",
```

I believe the `type` should be `object`.

gravesm:
Closes #496. This modifies the JSON schema to support geoblacklight
records containing a single layer and an array of layers.


drh-stanford:
Great! Thank you. 👍

PR #502, opened 2016-09-29
Data relation UI widget

sgbalogh:
Just picking this up from the end of the sprint... definitely not ready to merge, but thought it would be easier to discuss as a PR.

This implements some simple parent/child relation tracking in GeoBlacklight. It is a refactor from the version that I'm currently using for the NYU deployment. It creates two routes which return a corresponding record's 'ancestors' and 'descendants', where those are determined by Solr queries that look for a value(s) stored in another record's `dc_source_sm` field. Descendants need to point to their ancestor(s), but not vice versa.

The routes are:
```
localhost:3000/catalog/fake-record-001/ancestors.json
localhost:3000/catalog/fake-record-001/descendants.json
```
... and what is returned looks something like this:
```json
{
  "url": "http://localhost:3000/",
  "response": {
    "numFound": 2,
    "start": 0,
    "docs": [
      {
        "dc_title_s": "2016 NYC Geodatabase, ArcGIS Version (jan2016)",
        "layer_slug_s": "nyu_2451_34636"
      },
      {
        "dc_title_s": "2016 NYC Geodatabase, Open Source Version (jan2016)",
        "layer_slug_s": "nyu_2451_34635"
      }
    ]
  }
}
```

I also wrote a bit of JS to inject the results, returned to the browser via XHR, into a record view. That code is probably the messiest part of this right now.

Would love to hear any comments or recommendations about how to improve it. I realize everyone is probably busy on other projects currently, so no hurry!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,7 +25,7 @@ Metrics/AbcSize:
   Max: 23
 
 Metrics/MethodLength:
-  Max: 15
+  Max: 16
 
 # TODO: Add top-level class and module documentation.
 Style/Documentation:
--- a/app/assets/javascripts/geoblacklight/modules/relations.js
+++ b/app/assets/javascripts/geoblacklight/modules/relations.js
@@ -1,0 +1,15 @@
+Blacklight.onLoad(function () {
+    $('[data-relations="true"]').each(function (i, element) {
+        var $elem = $(element);
+        var attributes = $elem.data();
+        var relation_url = attributes['url'] + '/relations';
+        $.ajax({
+            url: relation_url,
+            type: 'GET',
+            dataType: 'html',
+            success: function (data) {
+                $elem.append($(data).hide().fadeIn(200));
+            }
+        })
+    });
+});
--- a/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
@@ -34,6 +34,14 @@
   @extend .fa, .fa-envelope;
 }
 
+.geoblacklight-relations-ancestor {
+  @extend .fa, .fa-pagelines;
+}
+
+.geoblacklight-relations-descendant {
+  @extend .fa, .fa-leaf;
+}
+
 .geoblacklight-web_services {
   @extend .fa, .fa-external-link;
 }
--- a/app/controllers/relation_controller.rb
+++ b/app/controllers/relation_controller.rb
@@ -1,0 +1,17 @@
+class RelationController < ApplicationController
+  include Blacklight::Configurable
+  include Blacklight::SearchHelper
+  copy_blacklight_config_from(CatalogController)
+
+  def relations
+    @relation_response = Geoblacklight::Relation::RelationResponse.new(params[:id], repository)
+    respond_to do |format|
+      format.json do
+        render json: @relation_response.relations
+      end
+      format.html do
+        render 'catalog/relations', layout: false
+      end
+    end
+  end
+end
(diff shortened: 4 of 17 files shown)


app/assets/javascripts/geoblacklight/modules/relations.js.erb
```diff
@@ -0,0 +1,80 @@
...
+var populateDescendantLinks = function (title, slug, baseurl) {
+    $('#relation_tools .nav').append('<li><a href="' + urlify_record(baseurl, slug) + '"><span class="geoblacklight geoblacklight-relations-descendant"> </span> ' + title + '</a></li>');
+}
+
+var populateDescendantSearchLink = function (ancestor_slug, count, baseurl) {
+    $('#relation_tools .nav').append('<li><a href="' + urlify_search(baseurl, ancestor_slug) + '">Browse all ' + count + ' items...</a></li>');
+}
+
+var urlify_search = function (site_url, slug) {
+    return site_url + '?f[<%= Settings.FIELDS.SOURCE %>][]=' + slug
```
mejackreed: Could this be settable as a data-attribute pattern rather than an erb template?

app/assets/javascripts/geoblacklight/modules/relations.js.erb
```diff
@@ -0,0 +1,80 @@
+var setupDataRelations = function () {
```
mejackreed: Can we remove the global functions here?

app/assets/javascripts/geoblacklight/modules/relations.js.erb
```diff
@@ -0,0 +1,80 @@
...
+}
+
+var urlify_search = function (site_url, slug) {
+    return site_url + '?f[<%= Settings.FIELDS.SOURCE %>][]=' + slug
+}
+var urlify_record = function (site_url, slug) {
+    return site_url + 'catalog/' + slug;
+}
+
+var grab_document_url = function () {
```
mejackreed: rather than having this get from the document, can we provide in a data-attribute?

app/assets/javascripts/geoblacklight/modules/relations.js.erb
```diff
@@ -0,0 +1,80 @@
+var setupDataRelations = function () {
+    $('#sidebar').append($('<div class="panel panel-default show-tools" id="relation_tools"><div class="panel-heading">Data Relations</div><div class="panel-body"><ul class="nav"><div class="ancestor_links"></div><div class="descendant_links"></div></ul></div></div>').hide().fadeIn(200));
+}
+
+var setupAncestorLinks = function (count) {
+    $('#relation_tools .nav').append('<b>Source Datasets (' + count + ')</b>');
+}
+
+var setupDescendantLinks = function (count) {
```
mejackreed: I'm wondering if we can't provide a route that renders these and just inserts the HTML? (rather than providing JavaScript templates)

sgbalogh:
Awesome, thanks @mejackreed. Would love to talk through it.

sgbalogh:
Another thing: do we want the `/ancestor.json` and `/descendant.json` endpoints to be responsive to any Blacklight API parameters? For instance, sending `rows=40` to the JSON view of the catalog affects the RSolr query, and I replicate that (or any other relevant params?) here if there would be a use for it.

sgbalogh:
Just a quick update on the last commit.

I tried implementing @mejackreed's suggestion about adding a route that renders the additional sidebar div as HTML. Now there are `/relations.html` and `relations.json` routes. On record page load, it makes a single XHR request to `relations.html` and appends it to the page.

There is also a JSON route which consolidates all of the ancestor/descendant info. Perhaps this makes the `/ancestors.json` and `/descendants.json` routes superfluous?

mejackreed:
Agree we can probably kill `/ancestors` and `/descendants` unless we come up with a need.

sgbalogh:
👍

sgbalogh:
Images of the widget:
child_view
parent_view

PR #506, opened 2016-10-18
Data relation UI widget

sgbalogh:
Creating a new PR since some refactoring went on since the last one...

@mejackreed was a major help in implementing this.

This PR provides simple parent/child relation tracking in GeoBlacklight. It creates one route (with HTML and JSON responses) which returns a corresponding record's 'ancestors' and 'descendants', where those are determined by Solr queries that look for a value(s) stored in another record's `dc_source_sm` field (or whatever other element is in `Settings.FIELDS.SOURCE`). Descendants need to point to their ancestor(s), but not vice versa.

The route is: `localhost:3000/catalog/fake-record-001/relations`

The HTML response returns a list that can be directly appended (via a XHR request) under the sidebar of a catalog#show page. A data content tag is added via the `_relations_container` partial to anchor the position for appending, and provide the current URL.  If there are no ancestors/descendants for a record, the response is an empty string.

A JSON response looks like this:
```json
{
  "url": "http://localhost:3000/",
  "response": {
    "numFound": 2,
    "start": 0,
    "docs": [
      {
        "dc_title_s": "2016 NYC Geodatabase, ArcGIS Version (jan2016)",
        "layer_slug_s": "nyu_2451_34636"
      },
      {
        "dc_title_s": "2016 NYC Geodatabase, Open Source Version (jan2016)",
        "layer_slug_s": "nyu_2451_34635"
      }
    ]
  }
}
```


app/views/catalog/relations.html.erb
```diff
@@ -0,0 +1,39 @@
...
+              <% @relation_response.relations[:descendants]['docs'][0..2].each do |descendant| %>
+                  <li>
+                    <%= link_to solr_document_path(descendant['layer_slug_s']) do %>
+                        <span class='geoblacklight geoblacklight-relations-descendant'></span> <%= descendant['dc_title_s'] %>
+                    <% end %>
+                  </li>
+              <% end %>
+              <% unless (@relation_response.relations[:descendants]['numFound'].to_i <= 3) %>
+                  <li>
+                    <%= link_to search_catalog_path({f: {"#{Settings.FIELDS.SOURCE}" => [@relation_response.relations[:current_doc]]}}) do %>
```
mejackreed: I've tried testing this and I could not get it to work.

lib/geoblacklight/relation/relation_response.rb
```diff
@@ -0,0 +1,20 @@
+module Geoblacklight
+  module Relation
+    class RelationResponse
+      attr_reader :relations
+      def initialize(id, repository)
+        @search_id = id
+        @repository = repository
+        @relations = {
+          ancestors: Geoblacklight::Relation::Ancestors.new(@search_id, @repository).results,
```
eliotjordan: Just noticed that this is refactored in PR #510.

PR #508, opened 2016-10-19
loosen up index view reset spec regex

mejackreed:
Though I wonder if `/^1 - \d of \d$.*/`is more flexible?


tampakis:
makes sense to me, it requires there to be at least one digit

PR #510, opened 2016-10-20
Modify routes to use exportable, and use rails rendering engine

mejackreed:
This is just a proposal, but I'm also thinking that we need some better tests around this new functionality


eliotjordan:
Using exportable seems like the right thing to do here. When the tests pass, I'm 👍

lib/geoblacklight/relation/relation_response.rb
```diff
@@ -1,19 +1,22 @@
...
-          current_doc: id
-        }
+      end
+
+      def ancestors
+        Geoblacklight::Relation::Ancestors.new(@search_id, @repository).results
+      end
+
+      def descendants
+        Geoblacklight::Relation::Descendants.new(@search_id, @repository).results
```
eliotjordan: We should memoize these reduce the calls to Solr. Right now, relation#index make 7 calls to solr. Two for ancestor and five for descendants.

` @descendants ||= Geoblacklight::Relation::Descendants.new(@search_id, @repository).results`

mejackreed: 👍  modified and pushed.

lib/geoblacklight/relation/relation_response.rb
```diff
@@ -1,19 +1,22 @@
 module Geoblacklight
   module Relation
     class RelationResponse
-      attr_reader :relations
+      attr_reader :id
```
mejackreed: Thanks good catch on that. Updated.

app/views/relation/index.html.erb
```diff
@@ -0,0 +1,38 @@
...
+          <% @relations.ancestors['docs'].each do |ancestor| %>
+            <li>
+              <%= link_to solr_document_path(ancestor['layer_slug_s']) do %>
+                <span class='geoblacklight geoblacklight-relations-ancestor'></span> <%= ancestor['dc_title_s'] %>
+              <% end %>
+            </li>
+          <% end %>
+        <% end %>
+        <% unless (@relations.descendants['numFound'].to_i == 0) %>
+          <b><%= t('geoblacklight.relations.descendant') %>(<%= @relations.descendants['numFound'] %>)</b>
```
sgbalogh: Could you add a space between the descendant string and the open paren? I think there was one before, maybe just obscured by the fact that this was previously spanning two lines.

mejackreed: Ah gotcha.. Just kicked this out to their own partial and i18n'ed the count.

PR #513, opened 2016-11-04
Adds tests for relation lib classes and relation controller

sgbalogh:


eliotjordan:
:+1: Nothing to add here. Looks good.

PR #521, opened 2017-01-06
use html translations for homepage

mejackreed:


tampakis:
Cool feature! I think these lines need to be updated to match the view partial change: https://github.com/geoblacklight/geoblacklight/blob/use-html-translations-for-homepage/config/locales/geoblacklight.en.yml#L10-L11

PR #529, opened 2017-02-09
allow blank search to retrieve full results set

### LINKED ISSUE #526: Empty search should fetch results page with all records
Clicking the search button on the homepage without entering a query causes the homepage to load again. Add the ability to have an "empty query" return the entire results set.

This behavior is currently active as a customization in EarthWorks: https://earthworks.stanford.edu/?utf8=%E2%9C%93&q=

tampakis:
Closes #526. Makes the homepage search form fields behave more consistently with search results page search form: https://github.com/projectblacklight/blacklight/blob/c5b170a0a404fd97df32f80dbc47fa58d3c9c152/app/views/catalog/_search_form.html.erb

By configuring a single search field in the catalog_controller, Blacklight will submit a url `search_field` parameter, which causes the `has_search_parameters?` to return true in https://github.com/geoblacklight/geoblacklight/blob/3603575c85b81b4070d56f3011dd9b55873a2697/app/views/catalog/index.html.erb#L1 (and render the search results).

### DIFF
--- a/app/views/catalog/_search_form_no_navbar.html.erb
+++ b/app/views/catalog/_search_form_no_navbar.html.erb
@@ -1,10 +1,12 @@
 <%= form_tag search_action_url, :method => :get, :class => 'search-query-form clearfix' do %>
 <%= render_hash_as_hidden_fields(search_state.params_for_search().except(:q, :search_field, :qt, :page, :utf8)) %>
 
-<% unless search_fields.empty? %>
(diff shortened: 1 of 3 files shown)


mejackreed:
This looks great @tampakis. Could this be considered a breaking change? (changing the default behavior) I'm inclined to say no, but wanted to just put the comment out there to see what others thought. 

Also, would mind adding a quick test for this?

PR #532, opened 2017-02-09
Remove solr_wrapper from geoblacklight generator

tampakis:
solr_wrapper is a dependency of Blacklight and gets installed as part of the Blacklight install generator. Removes redundant/unused install generator code for GeoBlacklight generator and template.


eliotjordan:
@mejackreed I don't remember either. Oh well. Seems to work just fine without it.

Issue #524, opened 2017-02-01
add support for geometry-based (i.e., polygon) bounding regions


drh-stanford:
Currently in 1.x GeoBlacklight only supports rectangular bounding boxes

mejackreed:
For searching, display (preview) or both?

drh-stanford:
more for display/preview, but maybe the API could search by polygon too

sgbalogh:
Was just talking about this with @eliotjordan actually –– it could be useful if the solr_geom for a record could more precisely reflect the physical geometry of the location by using a polygon, even if the search interface still issued rectangular search queries. It would fix the problem of, for instance, getting hits for USA records while looking at regions off the coast/across the border that still fall inside a bounding box of the USA (as is the case with large amounts of Canada or Mexico, etc.).

mejackreed:
Yep this would be great, but would complicate Solr setup as it would force users to manually install JTS.

sgbalogh:
Ahh, good point. Do you think that would be sufficiently annoying to justify leaving this out?

Just thinking out loud... another way to not cut users who do not have the ability to add JTS to their Solr env and still allow them to use shared OGM metadata with complex geoms would be to add a function in GeoCombine to extract rectangular bounding boxes and index those simple geoms instead. That would probably force a related (but perhaps smaller?) issue of requiring people to install GDAL...

mejackreed:
I think if we added smarter parsing of the geometry (like [blacklight-heatmaps](https://github.com/sul-dlss/blacklight_heatmaps/blob/master/app/models/concerns/blacklight_heatmaps/geometry_parser.rb) does) we could do this without having to add the solr requirement

sgbalogh:
Hm, not sure I understand. Isn't the above just code for parsing geo objects from different representations of them?

mejackreed:
Yep, so if we know how to parse things in a smart way.. we can index that field as anything SRPT https://cwiki.apache.org/confluence/display/solr/Spatial+Search#SpatialSearch-RPT . And people can index whatever they want. (they already can ... its just not really supported in the software at the moment)

sgbalogh:
Oh, right. But it would still be the case that if NYU is publishing records with complex polygon geoms and someone without JTS wanted to index our records, they wouldn't be able to. Thats what I was suggesting with a geometry simplifier component of GeoCombine, just as a way to make all records "backwards compatible" if we did this

[...]

PR #533, opened 2017-02-11
Complex geom

sgbalogh:
Addresses #524 –– adds ability to parse Solr ENVELOPE or WKT polygons from record into GeoJSON Leaflet layers in order to support complex display geometries.

Would be nice to have some tests focusing on the new functionality, but any fixtures with WKT polygons will require a Solr install with JTS to support. Perhaps a compromise would be just testing [these functions](https://github.com/geoblacklight/geoblacklight/blob/complex_geom/app/models/concerns/geoblacklight/solr_document/geometry.rb)?


app/assets/javascripts/geoblacklight/modules/results.js
```diff
@@ -1,76 +1,80 @@
-Blacklight.onLoad(function() {
-  var historySupported = !!(window.history && window.history.pushState);
+Blacklight.onLoad(function () {
+    var historySupported = !!(window.history && window.history.pushState);
```
mejackreed: Are these just whitespace changes? If so can we move these to separate commits? 

We standardize on two spaces everywhere, it looks like this is using 4 spaces? I realize that we could probably  update the tooling to better support this. https://github.com/geoblacklight/geoblacklight/blob/master/.jshintrc#L32

sgbalogh:
@mejackreed thanks for taking a look! I think you're right about the whitespace –– I'll look at that.

About the GeoJSON indexing –– I did experiment with that a little (though in general there is not much in the way of documentation on it). I noticed, though, that enabling it (which you have to do by adding a `format="GeoJSON"` tag to the spatial field in schema.xml) seems to force all geometries to be in GeoJSON, and (stranger) only responds to spatial queries that use GeoJSON. The ENVELOPE syntax is, I guess, just a flavor of WKT though –– so it seemed like sticking with WKT for complex polygons might be the best way not to break backward compatibility with records. But maybe there is a way around this that I just haven't seen documented.

And the feature is always active –– basically, if there a complex polygon in WKT as part of the spatial field in a record, it will now be displayable in the results page, or shown on a show page if applicable. Spatial queries are still sent as rectangles from the Leaflet viewport, but being able to index complex shapes helps immensely with spatial search for records with coverage as large as the USA, for example.

A potential risk would be in bombarding the DOM with a ton of complex GeoJSON layers on the results page, but I'm not convinced this would be a serious problem, especially since all layers are displayed one at a time.

mejackreed:
Gotcha about the GeoJSON.

I do have concerns about adding this much functionality without having end to end feature tests. I wonder if we could figure out to add JTS into solr wrapper for our test suite?

mejackreed:
Maybe https://github.com/cbeer/solr_wrapper/blob/master/lib/solr_wrapper/instance.rb#L30 this could help us?

sgbalogh:
I'm going to close this, since it seems related to issues we want to talk through more.

PR #537, opened 2017-04-19
Update results.js

ewlarson:
Add decimal support to parseableBbox RegExp; Remove fixed bbox default. Thanks and HT to @nihiliad

The B1G Geoportal bounding box data attributes have decimals, ex: "-97.991522 42.891013 -87.177541 49.892443".  As such, the 1.4.0 release parseableBbox RegExp pattern could not parse any of our bboxes.  Also reverting the bbox var instantiation so the bbox.extend conditional branch can work.

### DIFF
--- a/app/assets/javascripts/geoblacklight/viewers/map.js
+++ b/app/assets/javascripts/geoblacklight/viewers/map.js
@@ -9,7 +9,7 @@ GeoBlacklight.Viewer.Map = GeoBlacklight.Viewer.extend({
     bbox: [[-80, -195], [80, 185]],
     opacity: 0.75
   },
-  
+
   overlay: L.layerGroup(),
 
   load: function() {
@@ -17,6 +17,12 @@ GeoBlacklight.Viewer.Map = GeoBlacklight.Viewer.extend({
       this.options.bbox = L.bboxToBounds(this.data.mapBbox);
     }
     this.map = L.map(this.element).fitBounds(this.options.bbox);
+
+    // Add initial bbox to map element for easier testing
+    if (this.map.getBounds().isValid()) {
+      this.element.setAttribute('data-js-map-render-bbox', this.map.getBounds().toBBoxString());
+    }
+
     this.map.addLayer(this.selectBasemap());
     this.map.addLayer(this.overlay);
     if (this.data.map !== 'index') {
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,17 @@
+### Descriptive summary
+
+Include what version of GeoBlacklight related to this issue (1.3.0, master, etc.) if appropriate, and any relevant tracebacks if you're reporting a bug.
+
+### Expected behavior
+
+### Actual behavior
+
+### Steps to reproduce the behavior
+
+1. Do this
+1. Then do this...
+
+### Related work
+
+Link to related tickets or prior related work here.
+
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,7 +25,7 @@ Metrics/AbcSize:
   Max: 23
 
 Metrics/MethodLength:
-  Max: 15
+  Max: 16
 
 # TODO: Add top-level class and module documentation.
 Style/Documentation:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,21 +5,19 @@ notifications:
   email: false
 
 rvm:
-  - 2.3.1
+  - 2.3.4
 
 matrix:
   include:
-    - rvm: 2.3.1
-      env: "RAILS_VERSION=4.2.7.1"
-    - rvm: 2.2.5
-      env: "RAILS_VERSION=5.0.0.1"
+    - rvm: 2.3.4
+      env: "RAILS_VERSION=5.0.3"
 
 before_install:
   - gem update --system
   - gem install bundler
 
 env:
- - "RAILS_VERSION=5.0.0.1"
+ - "RAILS_VERSION=5.0.3"
 
 global_env:
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,16 +10,16 @@ rvm:
 matrix:
   include:
     - rvm: 2.3.1
-      env: "RAILS_VERSION=4.2.6"
+      env: "RAILS_VERSION=4.2.7.1"
     - rvm: 2.2.5
-      env: "RAILS_VERSION=5.0.0"
+      env: "RAILS_VERSION=5.0.0.1"
 
 before_install:
   - gem update --system
   - gem install bundler
 
 env:
- - "RAILS_VERSION=5.0.0"
+ - "RAILS_VERSION=5.0.0.1"
 
 global_env:
   - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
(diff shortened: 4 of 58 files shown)


mejackreed:
Thanks for the contribution @ewlarson . Do you have some test data where this fails? I wonder if it would be useful to add that to our fixtures so that we can write a test for this functionality to better support it?

ewlarson:
Sorry!  We do not have any test data/fixtures to reproduce the issue.  It's just something we noticed during our upgrade.  We really only have a small suite of capybara tests that ensure a group of exemplar records behave as expected.

I'll work on adding some tests to this PR.

ewlarson:
Okay, I think this is about as far as I can carry this PR.  I've updated the javascript code to add a data attribute to the div#map element that displays the initial bbox for the map view.  This allows for specs that query GBL and test whether that initial view bbox is accurate granted the bboxs for the docs that are returned in the result set.

The last hang up here seems related only to the Rails 4.2 build.  Anyone have any insights to share?

app/assets/javascripts/geoblacklight/viewers/map.js
```diff
@@ -9,14 +9,20 @@ GeoBlacklight.Viewer.Map = GeoBlacklight.Viewer.extend({
...
+
   overlay: L.layerGroup(),
 
   load: function() {
     if (this.data.mapBbox) {
       this.options.bbox = L.bboxToBounds(this.data.mapBbox);
     }
     this.map = L.map(this.element).fitBounds(this.options.bbox);
+
+    // Add initial bbox to map element for easier testing
```
mejackreed: I'm a little unsure here why we need to set this. Shouldn't this be coming from the view already?

sgbalogh:
I tested this out and it works great. @ewlarson, do you know if that conditional in map.js that Jack pointed out is necessary? Would love to get this PR merged in!

sgbalogh:
Hmmm... actually, it appears like there is a race condition somewhere. The Leaflet map is trying to be instantiated twice on results pages. I'll dig into this more...

mejackreed:
@sgbalogh That may have to do with https://github.com/geoblacklight/geoblacklight/pull/537/files#diff-532493f9653f312ca97046bddc0064c7R21

I'm a little uneasy about that block

ewlarson:
Hi all, thanks for testing it.  I'll be back in the office next Tuesday and
can take a look.  In general, I was trying to output the final bbox "view"
of the result's map into a data attribute, which could be used during
testing to ensure that initial bbox was sensible.  I don't think that value
appears anywhere else on the page, and it's calculated during the map init.

sgbalogh:
@ewlarson thanks for the clarification! And sorry to bother with this while you're out of the office.

I think what's going on is, (re @mejackreed's comment), the Leaflet map will have a 'data-map-bbox' attribute that is set by the view only in situations where there is a param for bbox (see [here](https://github.com/geoblacklight/geoblacklight/blob/3603575c85b81b4070d56f3011dd9b55873a2697/app/views/catalog/_document_split.html.erb#L4)). Otherwise it's null. Therefore, the document split map only has this attribute when people are doing spatial search.

If that results map pops up in response to any non-spatial query, the bbox is determined entirely at the client side by that results.js code. Eric's code just injects the client-computed bbox into the data attribute for consistency / testing.

And my comment about a race condition earlier I think was my mistake... my attempt to override some related JS functions just backfired during testing.

mejackreed:
@tampakis Do you have any thoughts here? I know you just worked on #520 which had some similar issues.

ewlarson:
Thanks @mejackreed and @sgbalogh for the detailed feedback.

Ahhh... I'm sorry for the confusion here, since I've accidentally created a data attribute (data-map-bbox) naming collision with my code in this PR.

The view only adds the data-map-bbox attribute when a bbox param exists, but many searches (ex. clicking a homepage facet) do not cause the attribute to be written.  The code here adds the data-map-bbox attribute anytime the map.js instance has parsable div.row[@bbox] values.

If I rename my data attribute to data-map-initial-bbox I think that might eliminate the confusion and help push this PR through.  Sound okay?

Also, @sgbalogh the map double-init render/race condition isn't a thing, right?

sgbalogh:
@ewlarson correct, not a thing! Sorry about the confusion regarding that. Thanks for taking a look.

ewlarson:
ALSO!  Is there a place where the view's data-map-bbox attribute is tested/used by the application?  Another direction to consider here would be to remove the html view's generation of the bbox attribute, and always append it via JS.

mejackreed:
Gotcha. I understand where you are coming from @ewlarson . I'm still a bit concerned about updating that data attribute property from the JavaScript just for testing sake. My concern really isn't on your implementation but potentially causing confusion down the road or causing side affects with how those attributes are set.

What do you think? I might just be overly cautious here and am definitely open to being convinced otherwise :)

ewlarson:
Okok... an attempt to convince :)

GBL releases 1.4.0 to current all have the [bbox regex parsing bug](https://github.com/BTAA-Geospatial-Data-Project/geoportal/issues/27) we encountered upon upgrading the B1G Geoportal to 1.4.0.  This PR fixes that issue at the expense of populating a data attribute (which already exists in GBL sometimes), and will help us avoid any future regression.  Bug fixed.

And I _absolutely agree_ it's fair to question why and how map data attributes are generated and populated.  However, the URL bbox param value gets shuttled straight into the data-map-bbox attribute, but it's NOT the actual bbox the map view renders at -- that absolute value is only accessible via calling the JS, this.map.getBounds().toBBoxString(), which my code does.  Many times (always?) the search constraint and the map bbox are very slightly different.

Or, try manipulating your bbox param value by hand to something dumb like: http://localhost:3000/catalog?bbox=neo%20hoodoo%20voodoo and well... that's not a valid parsable bbox but it's now a map data attribute which smells like well... hoodoo.  Kinda odd.

My favorite fix here is to eliminate the data attribute generation from the html view (again, that value is not reused or tested within the app currently) in favor of adding the bbox data attribute from JS which now has tests with this PR.  Let's lean heavily towards eliminating the bug.

Talk it over during the call tomorrow?

mejackreed:
Sounds good! I think I'm convinced :) Mind squashing stuff down?

ewlarson:
Yup, on it.

mejackreed:
Ah.. maybe lets talk it over on call. I think I misunderstood you. I was ok with the JS approach but removing from the data attributes I'm not so sure of.

ewlarson:
Sure thing

ewlarson:
Well, looks like I've squashed things here rather poorly.

Instead of 1 commit I see 15 (gah!), with [one commit included](https://github.com/geoblacklight/geoblacklight/pull/537/commits/8f6204ae1c2e22c4b2b9d72872a9a8f5bcab29d3) that has the prior commits squashed.  I guess forgetting to push using force made this happen?

Now, subsequent attempts to squash commits including the pre-squashed commit aren't working.  Anyone have any hints or suggestions?  Feel like the easiest path here might be to redo this PR...

sgbalogh:
Oh, I think you do need to do a force push after the squash. I think I've followed [this](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) before. So, basically use the interactive rebase to select the commits you want to squash down, verify in your git log that it looks the way you want it to, then force push it back into your branch. I think that should do it?

ewlarson:
Sorry!  Going to retry things in a new PR.  Can't make any sense of my busted squash within a squash.

PR #539, opened 2017-04-28
Bugfix for getFeatureInfo responses with multiple rows

### LINKED ISSUE #538: Geoblacklight::FeatureInfoResponse bug when WMS response contains multiple HTML rows
I ran into this small bug on 1.4.0:

When trying to visualize info for a WMS feature, `FeatureInfoResponse` throws an error whenever multiple HTML rows of features are returned from the WMS service.

By default this never happens, since a request to a WMS service that does not specify a `FEATURE_COUNT` param should always receive only one feature in response. For a different reason, I needed to explicitly provide a FEATURE_COUNT > 1 (in Geoblacklight::Settings) and request multiple features at a time.

Our UI for displaying feature attributes is only capable of displaying one attributes for one feature at a time. I expected the behavior to automatically just select the first returned feature from the WMS response, but whenever there are multiple HTML rows it crashes. See https://github.com/geoblacklight/geoblacklight/blob/v1.4.0/lib/geoblacklight/wms_layer/feature_info_response.rb#L15-L25

I'm pushing a very small tweak to this code and an accompanying test.

sgbalogh:
Fixes #538

### DIFF
--- a/lib/geoblacklight/wms_layer/feature_info_response.rb
+++ b/lib/geoblacklight/wms_layer/feature_info_response.rb
@@ -19,7 +19,7 @@ module Geoblacklight
         table_values[:values].push([th.text])
       end
       page.css('td').each_with_index do |td, index|
-        table_values[:values][index].push(td.text)
+        table_values[:values][index].push(td.text) unless index >= table_values[:values].count
       end
       table_values
     end
--- a/spec/lib/geoblacklight/wms_layer/feature_info_response_spec.rb
+++ b/spec/lib/geoblacklight/wms_layer/feature_info_response_spec.rb
@@ -2,6 +2,7 @@ require 'spec_helper'
 
 describe Geoblacklight::FeatureInfoResponse do
   let(:response) { described_class.new(OpenStruct.new(body: '<html><body><table><th>Header1</th><th>Header2</th><td>value1</td><td>value2</td></body></html>', headers: { 'content-type' => 'all good' })) }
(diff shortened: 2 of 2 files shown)


mejackreed:
Looks good to me. Is it worth adding a fixture for this functionality?

sgbalogh:
@mejackreed: Just to clarify –– do you mean turning the lines at https://github.com/geoblacklight/geoblacklight/blob/fb81659741d66a19b3e9b1c958ab7505022b5ba6/spec/lib/geoblacklight/wms_layer/feature_info_response_spec.rb#L4-L5 into fixtures that store complete responses from a WMS endpoint?

mejackreed:
I was thinking more about adding a representative dataset (document) into here: https://github.com/geoblacklight/geoblacklight/tree/fb81659741d66a19b3e9b1c958ab7505022b5ba6/spec/fixtures/solr_documents

For instance, how are you turning on this functionality? Are you sending in multiple layer names?

sgbalogh:
@mejackreed: Oh, gotcha.

No, the impetus was for a much stupider reason than anything like that. I want an index map from a WMS preview of a single layer that is a MultiPolygon type, and to have overlapping features be queryable –– for example, if a smaller polygon runs over a larger polygon in the background, clicking on an area of the smaller polygon should return that feature's info –– otherwise it should return the info for the larger background polygon. For whatever reason, GeoServer by default returns polygons in order of drawing (I think that's the logic) for GeoFeatureInfo –– meaning that the bottom-most features always come first –– and I actually want the reverse of that...

Long story short, it's possible to flip the rendering of the features in GeoServer by overriding a template, but only if you request more than 1 feature at a time via WMS... hence the problem.

So this would require no changes to the actual Solr record –– it's still just a single layer (just with multiple features at various x,y points).

mejackreed:
Gotcha thanks, yeah that makes sense I just didn't understand. 🆒

PR #542, opened 2017-05-24
add iiif_manifest as external service type

tampakis:
Adds support for a IIIF manifest type in the external services reference hash.

### DIFF
--- a/lib/geoblacklight/constants.rb
+++ b/lib/geoblacklight/constants.rb
@@ -7,6 +7,7 @@ module Geoblacklight
       geojson: 'http://geojson.org/geojson-spec.html',
       html: 'http://www.w3.org/1999/xhtml',
       iiif: 'http://iiif.io/api/image',
+      iiif_manifest: 'http://iiif.io/api/presentation#manifest',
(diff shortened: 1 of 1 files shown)


ewlarson:
Just saw that your PR 4.2 build fails like mine.  Looks like an error with the ActiveRecord::Migration[4.2] version statement, which exists in 5+ but doesn't actually exist in 4.2?
http://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html

mejackreed:
Is this a proposal to start supporting manifests in DCT references?

tampakis:
@mejackreed, it sure is

mejackreed:
I'm curious then if that is a proposal, should we look at adding this in documentation and what the potential implications are for viewing the data?

eliotjordan:
@tampakis is putting together a PR for Pulmap that uses an embedded Universal Viewer to visualize the manifest. We could put together documentation on how to do something implement that in a local GeoBlacklight instance.

https://github.com/pulibrary/pulmap/compare/uv-display

PR #549, opened 2017-07-06
Update parseableBbox and add specs to validate map initial view

ewlarson:
Add decimal support to parseableBbox RegExp; Thanks and HT to @nihiliad

The B1G Geoportal bounding box data attributes include decimals, ex:
"-97.991522 42.891013 -87.177541 49.892443".  As such, the 1.4.0
release parseableBbox RegExp pattern could not parse any of our bboxes.

GeoBlacklight.Viewer.Map now adds a data-js-map-render-bbox attribute
upon valid bboxes, so specs can verify search result sets render an
appropriate initial map view bbox.


ewlarson:
Clean version of [the old PR](https://github.com/geoblacklight/geoblacklight/pull/537)

PR #553, opened 2017-07-25
[WAITING for 2.0] Leaflet upgrade

ewlarson:
Bump leaflet, esri-leaflet, and leaflet-iiif to their most current release versions.  Fix upgrade issues for the opacity slider, and make small leaflet api adjustments for esri viewer.  Adds an explicit skip auth verification to the WmsController, to avoid CSRF errors raised in Rails 5.1 (local testing version).


app/assets/javascripts/geoblacklight/viewers/esri.js
```diff
@@ -21,7 +21,7 @@ GeoBlacklight.Viewer.Esri = GeoBlacklight.Viewer.Map.extend({
     // remove any trailing slash from endpoint url
     _this.data.url = _this.data.url.replace(/\/$/, '');
 
-    L.esri.get = L.esri.Request.get.JSONP;
+    // L.esri.get = L.esri.Request.get.JSONP;
```
eliotjordan: I think it's OK to remove this line entirely. Documentation on the ESRI Leaflet tool says that fallback to JSONP is handled automatically now.

app/controllers/wms_controller.rb
```diff
@@ -1,4 +1,6 @@
 class WmsController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
mejackreed: I'm a little uneasy about this line. Is there a change in Leaflet that requires this or is this a Rails thing?

mejackreed:
Great work here! One quick comment/question really.

Another thing I was thinking, does this constitute a breaking change for some? I wonder if this is backwards compatible, and would others have issues with a heavily customized frontend. Perhaps Cornell, Princeton, VecNet?

cc @eliotjordan @dbrower @kgjenkins @starsplatter

dbranchini:
Thank you @mejackreed for thinking of us! I don't think this change will break anything on our site.

app/assets/javascripts/geoblacklight/viewers/esri.js
ewlarson: Yes, nice catch.  I had intended to remove all of the commented-out code.

dbrower:
@mejackreed This won't affect the VecNet site. We're currently locked on old version of geoblacklight. Thanks for the consideration.

app/controllers/wms_controller.rb
ewlarson: I had to add this to get Rails 5.1 to stop complaining about CSRF issues locally (no auth token).  Specifically, this was for the dynamic maps where points/areas dynamically populate the table.

So, yeah a Rails thing.  Not leaflet.  I figured since the controller only exposes the R in CRUD, it wasn't a big deal skip validation and make Rails happy.

eliotjordan:
@mejackreed This would be a breaking change for us, but we could get it integrated fairly quickly.

mejackreed:
Gotcha. So I guess one thing we need to consider is for GeoBlacklight how do we semantically version the software? What constitutes a breaking change (metadata schema change, ruby api change, ui change, javascript change?) and how can we integrate these things together to release software effectively that doesn't put adopters in a bad state.

ewlarson:
I think the semantic versioning discussion is bigger than this PR.  These javascript updates are backwards compatible, so they'd constitute a MINOR release.  Only adopters who have customized their map scripts will see any (potential) issues.

ALSO! I think @mejackreed is hinting at improved javascript dependency management in GBL.  In a future Rails 5.1 release, we should adopt yarn (package.json) to better articulate our JS dependencies.

mejackreed:
If the changes are backwards compatible, then I'm a big :+1: . Thanks @ewlarson

ewlarson:
The JS libraries in this older PR have aged rapidly:
* leaflet 1.1.0 is now 1.3.3
* esri-leaflet 2.0.8 is now 2.2.1
* leaflet-iiif 1.2.1 is now 2.0.1

Maybe it's worth re-upgrading before merging?

eliotjordan:
@ewlarson That seems like a good idea. Get everything up to the most recent version.

ewlarson:
Closing this PR. Superseded by PR #676

Issue #380, opened 2016-01-27
Direct download of metadata from metadata preview modal


eliotjordan:
Allow users to download FDGC and ISO metadata directly from metadata preview window. @sgbalogh

KarenMajewicz:
Completed with issue #555

PR #555, opened 2017-07-27
Displays transformed geospatial metadata in the HTML

### LINKED ISSUE #173: View and download metadata design

jrgriffiniii:
Resolves #173, #380, and #559 by addressing the following:
- Implements Classes for the transformation of geospatial metadata into the HTML
- Provides XSL Stylesheets for the transform

Still being addressed:
- [ ] ~~Migrating the XSL Stylesheet customizations into [GeoCombine](https://github.com/OpenGeoMetadata/GeoCombine)~~ (Postponed: https://github.com/OpenGeoMetadata/GeoCombine/issues/18#issuecomment-319491506)
- [x] Enabling downloads for the XML metadata
- [x] Styling for the metadata interface
- [x] Refactoring the `BaseTransformer` Class
- [x] Improve the inline documentation for `BaseTransformer`, `FcgdTransformer`, and `Iso19139Transformer`
- [x] Extend the Modal for missing metadata resources (e. g. 404 responses)


mejackreed:
A lot of these transforms are already implemented in the [GeoCombine gem](https://github.com/OpenGeoMetadata/GeoCombine)

Would it be possible to use those api's already developed and include these as a gem? (instead of maintaining both separately)

jrgriffiniii:
> Would it be possible to use those api's already developed and include these as a gem? (instead of maintaining both separately)

@mejackreed thank you very much for this review and guidance.  I shall proceed by focusing next upon [GeoCombine](https://github.com/OpenGeoMetadata/GeoCombine), and trying to migrate these XSL Stylesheets upstream.

app/assets/javascripts/geoblacklight/modules/metadata.js
```diff
@@ -0,0 +1,16 @@
+
+var setRefUrl = function setRefUrl() {
```
mejackreed: Could this be implemented in a way where a global variable function isn't added? I'm concerned with adding to the global scope here.

jrgriffiniii: Certainly.  0447abe refactors this as `GeoBlacklight.MetadataDownloadButton`, and provides some basic `teaspoon` tests.

jrgriffiniii:
@mejackreed I've refactored the jQuery functionality a bit further: https://github.com/geoblacklight/geoblacklight/pull/555/files#diff-10a899d9a4b773ff0cd7620459ec56d1
...as well as resolved a styling issue for the metadata Modal for narrower viewports, and removed CodeRay.

app/assets/stylesheets/geoblacklight/modules/metadata.scss
```diff
@@ -1,5 +1,28 @@
...
+
+.modal-header {
+  .modal-title {
+    font-weight: 400;
+    font-size: 20px;
+  }
+}
+
+.modal-body {
+  overflow-y: scroll;
```
eliotjordan: What do you think about constraining the body height so that that the entire modal fits on the screen? It would be nice to see that download is an option without scrolling to the bottom of the page. Here's what it looks like when you style  `modal-body` with `height: calc(100vh - 190px)`:
screenshot 2017-08-09 00 06 44

jrgriffiniii:
Thank you very much for reviewing this @eliotjordan , I believe that these improvements have been addressed.

PR #561, opened 2017-08-08
make bounding box constraint label configurable

tampakis:
This shouldn't be a breaking change and makes it easier to configure a custom label for the bounding box search constraint.


mejackreed:
👏

PR #563, opened 2017-08-10
Fix functionality for validating bounding boxes

### LINKED ISSUE #562: Issues with incorrect bounding box formatting
In the most recent version on GeoBlacklight (v1.5.1), there are two bugs related to bounds and bounding box formatting.

1. The [L.bboxToBounds](https://github.com/geoblacklight/geoblacklight/blob/master/app/assets/javascripts/geoblacklight/geoblacklight.js#L15) function returns incorrect bounds from properly formatted bbox strings. Leaflet defines a bounds array as `[[lat, lng],[lat, lng]]`, whereas this function returns `[[lng, lat],[lng, lat]]`.

2. The regex for determining parseable bounding boxes in [results.js](https://github.com/geoblacklight/geoblacklight/blob/master/app/assets/javascripts/geoblacklight/modules/results.js#L16-L17) is incorrect. Lat and Lng are inverted here; perhaps as a response to issue with issue 1 above. Longitude should have have a maximum of 3 digits before the decimal point, while longitude should have 2. This is the correct regex:

 ```javascript
  var lngRe = '(-?[0-9]{1,3}(\\.[0-9]+)?)';
  var latRe = '(-?[0-9]{1,2}(\\.[0-9]+)?)';
 ```

eliotjordan:
Closes #562.

### DIFF
--- a/app/assets/javascripts/geoblacklight/modules/results.js
+++ b/app/assets/javascripts/geoblacklight/modules/results.js
@@ -13,23 +13,34 @@ Blacklight.onLoad(function() {
     opts = { baseUrl: data.catalogPath },
     geoblacklight, bbox;
 
-    var lngRe = '(-?[0-9]{1,2}(\\.[0-9]+)?)';
-    var latRe = '(-?[0-9]{1,3}(\\.[0-9]+)?)';
+    validateBounds = function(bounds) {
+      if (bounds) {
+        return L.latLngBounds([[-90, -180],[90, 180]]).contains(bounds);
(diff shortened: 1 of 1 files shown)


eliotjordan:
@gravesm Is this what you had in mind?

gravesm:
What if we cleaned up the return value of the bboxToBounds function like this? This would make the flow control easier to follow.

mejackreed:
Is it possible to get a failing test for the previously incorrect behavior?

PR #567, opened 2017-08-23
Handle geospatial metadata as XML (without HTML transforms)

### LINKED ISSUE #566: Handle viewing of metadata that isn't converted
### Descriptive summary

After merge of #555, previously displayed metadata that does not have a conversion in GeoCombine, will cause page errors and not display. The error message seems incorrect as well.

screen shot 2017-08-21 at 3 57 10 pm

### Expected behavior
If there is no HTML conversion, the raw metadata should be shown.

### Actual behavior
The page reports that metadata is not available.

1. Navigate to http://127.0.0.1:3000/catalog/stanford-cg357zz0321/metadata
1. Try clicking on the "ISO 19139" button.

### Related work

#555 PR, and related issue #559

jrgriffiniii:
Resolves #566 by addressing the following:
- Handling cases in which metadata can be rendered as 
XML (using CodeRay)
- Ensures that metadata retrieved using the HTTP can be offered by endpoints issuing redirect responses
- Reintroducing core support for the MODS as a metadata type

Please find the attached screenshot for the interface updates:
geoblacklight_issues_566_screenshot

### DIFF
--- a/app/views/catalog/metadata/_markup.html.erb
+++ b/app/views/catalog/metadata/_markup.html.erb
@@ -1,0 +1,8 @@
+<div id="metadata-markup-container" class="container-fluid">
+  <div class="alert alert-warning" role="alert">
+    <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span>
+    <span class="sr-only">Error:</span>
+    <span>The metadata for this item cannot be transformed.</span>
+  </div>
+  <%= CodeRay.scan(content, :xml).div.html_safe %>
+</div>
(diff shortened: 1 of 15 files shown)


app/views/catalog/metadata/_markup.html.erb
```diff
@@ -0,0 +1,9 @@
+<div id="metadata-markup-container" class="container-fluid">
+  <!--<div class="well well-lg bg-warning">The metadata for this item cannot be transformed.</div>-->
```
eliotjordan: Remove commented out div.

Issue #569, opened 2017-10-24
Failing test suite


ewlarson:
### Descriptive summary

Unrelated to any specific GBL commit/code, the [test suite is failing](https://travis-ci.org/geoblacklight/geoblacklight/builds/289661635). It appears to be related to capybara/selenium element visibility issues.

### Expected behavior

I expect the master branch test suite to pass.

### Actual behavior

The master branch test suite fails, with several "Unable to find visible css" errors.

### Steps to reproduce the behavior

clone the repo
bundle exec rake ci

ewlarson:
Debugging notes: Some errors are rails_ujs related, too -- which was added to Rails 5.1.  The most recent BL releases slip us into Rails 5.1+ and the test suite fails.  Locking GBL down to BL 6.11.0 seems to solve things...

mejackreed:
This might be related, but I had to add some compatibility layer in Spotlight: https://github.com/projectblacklight/spotlight/pull/1834/files

I also remember an issue with CSRF needing to be removed or something? This may help.

ewlarson:
Ok, these tests largely fail because the catalog#show tool links -- Metadata, Cite, etc. -- open a modal that insists on re-fetching all the JS assets (layout wrapper, too).  BL 6.11.0 is the last release that doesn't exhibit the bug.

I'll spend some more time chasing down what happened between 6.11.0 and 6.11.1 to see what we can do to solve the issue.

mejackreed:
Maybe checkout this? https://github.com/projectblacklight/blacklight/pull/1797

ewlarson:
[That PR](projectblacklight/blacklight#1797) does fix things for email, sms, cite... but it doesn't solve the issue for our custom "metadata" action.

ewlarson:
It may be pinning ourselves to 6.11.0 like Avalon for the time being:
https://github.com/avalonmediasystem/avalon/pull/2395

PR #571, opened 2017-10-26
Wrap descendant and ancestor labels in li tags

### LINKED ISSUE #570: Wrap "descendant" and "ancestor" labels in <li> tags
### Descriptive summary

In GeoBlacklight 1.6 (and previous) the `descendant` and `ancestor` labels  in the descendants and ancestors partials should be wrapped in a `<li>` tag. `<b>` tags in unordered lists are not legal and can cause problems with accessibility.

eliotjordan:
Closes #570

### DIFF
--- a/app/views/relation/_ancestors.html.erb
+++ b/app/views/relation/_ancestors.html.erb
@@ -1,4 +1,6 @@
-<b><%= t('geoblacklight.relations.ancestor') %></b>
+<li>
+  <b><%= t('geoblacklight.relations.ancestor') %></b>
+</li>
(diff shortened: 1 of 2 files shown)


eliotjordan:
Travis is fails due to capybara issues. Probably related to: https://github.com/geoblacklight/geoblacklight/issues/569

mejackreed:
Do these also need to be wrapped up in a `<ul>` or is that already there?

eliotjordan:
The <ul> tag is already there.

Issue #574, opened 2017-12-13
Opensearch Support Broken


ewlarson:
### Descriptive summary

OpenSearch support is broken in master and several past releases. Appears to be at least a routing bug. GBL appears to have no routing specs itself?

### Expected behavior

Expect /catalog/opensearch.xml to render via: CatalogController#opensearch

### Actual behavior

Actual /catalog/opensearch.xml renders via: CatalogController#show as XML

The show action in turn queries Solr for a corresponding document and fails to find one, ultimately raising a Blacklight::Exceptions::RecordNotFound error. 

### Steps to reproduce the behavior

1. Launch Geoblacklight
2. Hit the /catalog/opensearch.xml route / [B1G Geoportal example](https://geo.btaa.org/catalog/opensearch.xml)

mejackreed:
I wonder if this is also an upstream Blacklight thing, or if we just have something configured incorrectly. https://demo.projectblacklight.org/catalog/opensearch.xml

ewlarson:
I think it might just be a route priority thing. I see Princeton has it right here: https://maps.princeton.edu/catalog/opensearch.xml

ewlarson:
Yeah that is indeed the issue. The fix requires getting BL's catalog#opensearch route above catalog#show in routes.rb

I'll cut a branch and add some basic routing specs to GBL, then I'll see if I can lift those routes up the chain.

mejackreed:
Resolved in #576

PR #576, opened 2017-12-13
Fix OpenSearch Route Priority

ewlarson:
Preserve OpenSearch route priority, by adding GBL routes after BL routes. This PR fixes the OpenSearch routing issue #574


ewlarson:
This PR will pass the test suite, after [the other one](https://github.com/geoblacklight/geoblacklight/pull/575) is accepted. :)

Issue #578, opened 2017-12-14
Update for Blacklight 7 deprecations


mejackreed:
- [x] render_document_show_field_value is deprecated and will be removed from Blacklight version 7.0.0 (replaced by ShowPresenter#field_value). See #583
 - [x] BlacklightHelper#application_name will no longer delegate to config.application_name in version 7.0. Set the i18n for blacklight.application_name instead.

eliotjordan:
Closed by #657

PR #583, opened 2018-01-09
migrate away from deprecated Blacklight helper method in favor of pre…

mejackreed:
…senter

Partial #578 

Does this need to wait for 2.0?


eliotjordan:
I think this is OK to include in a pre 2.0 release. As long as there is an upgrade note for those who've 
overridden the `upper_metadata` partial. We have in Pulmap, but we are already using the presenter. Both the helper and the presenter method will work until we upgrade to BL 7.

PR #589, opened 2018-01-17
add "more" link to icon facet partial

### LINKED ISSUE #587: Icon facet "More" not displaying
### Descriptive summary

The `_icon_facet.html.erb` partial is used for displaying facet lists for facets that have icons (institutions, data type). But the "more" link is not displaying for when there is a limit for the catalog controller.
### Expected behavior
The "More" link should be available.
### Actual behavior
No "more" link

screen shot 2018-01-16 at 10 28 54 am

mejackreed:
Closes #587

### DIFF
--- a/app/views/catalog/_icon_facet.html.erb
+++ b/app/views/catalog/_icon_facet.html.erb
@@ -1,8 +1,15 @@
 <ul class="facet-values list-unstyled">
   <% paginator = facet_paginator(facet_field, display_facet) %>
   <% paginator.items.each do |item| -%>
-  <li>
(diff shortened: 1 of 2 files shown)


ewlarson:
This works well. Just to be thorough, let's toss a quick spec in here, too. After this line:
https://github.com/geoblacklight/geoblacklight/blob/083dc61fd4a117f0c15a518a6e73e81d868a9bac/spec/features/split_view.html.erb_spec.rb#L30

Can you please add this?
expect(page).to have_css('a.more_facets_link', text: 'More', visible: true)

PR #633, opened 2018-07-25
Restructures the Downloads within the Tools panel

### LINKED ISSUE #607: Download panel refactoring
Open up the download panel so it is more configurable for different types of download presentations.

Inspired by Cornell's change to the download panel.

jrgriffiniii:
Resolves #607 by addressing the following:
- Removing the Downloads interface from the show_tools template partial
- Restructuring the templates into the partial "show_downloads" (which rebrands generated downloads as "exports")
- Adjusts the styling by introducing a "downloads" SCSS Module
- Adjusts the downloads JavaScript behavior in response to changes in the DOM
- Extends the locale in order to provide labels for secondary/generated downloads as "exports"


eliotjordan:
screenshot 2018-07-26 11 31 30


There are some file format i18n translation errors popping up on some of the file formats. We should probably add `esri_geodatabase` and `sqlite_database` to the formats list:

https://github.com/geoblacklight/geoblacklight/blob/master/config/locales/geoblacklight.en.yml#L18

Also, we'll need to add something like `format.downcase.gsub(" ", "_")` to the [proper_case_format](https://github.com/geoblacklight/geoblacklight/blob/master/app/helpers/geoblacklight_helper.rb#L65) helper method.

app/assets/stylesheets/geoblacklight/modules/downloads.scss
```diff
@@ -0,0 +1,82 @@
+%downloads-exports {
```
eliotjordan: Fancy Sass! No comment other than TIL about placeholder selectors.

mejackreed: Whoa, nice!

PR #652, opened 2018-07-26
639 search results formatting

### LINKED ISSUE #639: Search result formatting changes
Search result formatting is different:

- Icons are on the left instead of right (we should flip back).
- Text is very large (see other font issue), which causes crowding in the search area.
- No sorting and per page buttons

screenshot 2018-07-25 18 14 54

Versus

screenshot 2018-07-25 18 15 07

ewlarson:
Fixes #639 / Re-adds the sort and per_page widgets. Reorders .document child elements via flexbox's order rule. Tightens BL7 default spacing for facets and .document results, returning these GBL elements to a half-rem for spacing instead of a full rem.

### DIFF
--- a/app/assets/stylesheets/geoblacklight/_blacklight_overrides.scss
+++ b/app/assets/stylesheets/geoblacklight/_blacklight_overrides.scss
@@ -17,3 +17,18 @@
   position: relative;
 }
 // End - Addresses GBL issue #634
+
+// Begin - Addresses GBL issue #639
+.facet-limit {
+  margin-bottom:0;
+}
+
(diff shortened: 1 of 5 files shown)


ewlarson:
screen shot 2018-07-26 at 4 34 28 pm

ewlarson:
Nice catch! I'll take a further look...

ewlarson:
BL7 shows the exact same UI issue. I'll add an issue to BL and try to develop a fix. Want to merge this PR?

screen shot 2018-07-27 at 10 01 13 am

PR #668, opened 2018-07-30
Remove table-responsive class to fix issue with attribute column width

### LINKED ISSUE #645: Feature inspection table
Table for feature inspection has strange formatting after a feature is selected.

screenshot 2018-07-25 18 49 09

Versus...

screenshot 2018-07-25 18 49 14

eliotjordan:
Fixes issue with improperly sized feature inspection table by removing the `table-responsive` class. The table is not responsive, so it isn't needed.

Closes #645


ewlarson:
Did we want to follow the "Always responsive" Bootstrap 4 pattern here? Wrapping table.table inside a div.table-responsive:
https://getbootstrap.com/docs/4.1/content/tables/#always-responsive

PR #682, opened 2018-08-01
[WIP] fixes flipping bbox

christinach:
I would like to test some more fixtures before removing the WIP


christinach:
The PR fixes the flipping box because of coordinates crossing the antimeridian but will not fix coords that are passed incorrectly in the solr_geom field.

christinach:
@KarenMajewicz PR will update the solr_geom for https://github.com/geoblacklight/geoblacklight/pull/697/files#diff-07cc93333019091171bf8ad6d9894911. With the updated fixture it will look like:
2014newyorkcityplaces

christinach:
The show page for the area of Alaska that crosses the antimeridian because of the "solr_geom": "ENVELOPE(-179.9999845, 179.9993566, 71.9545187, 43.0532272)", 
will now display as:
100meterresolutionalaska360

eliotjordan:
@christinach This could use a rebase with master. And perhaps squashing those three commits into one?

Issue #317, opened 2015-04-04
structured data missing from documents in search results


mejackreed:
add schema.org microdata to search results (titles, year)

PR #709, opened 2018-08-03
[WIP] Add support for additional schema.org props

ewlarson:
Partially addresses #317

Adds name, temporal, author, and publisher to search results following the GBL/BL show page itemprop pattern. Needs further discussion to determine the correct set of props to include before closing this issue.

Some BL/GBL helpers, like snippit, should be refactored to wrap semantic itemprops around field data.


ewlarson:
Not much momentum here. Closing. Can revisit in the future.

PR #713, opened 2018-08-07
use a backwards compatible ujs adapter

### LINKED ISSUE #712: ReferenceError: Rails is not defined in WMS javascript
### Descriptive summary

in v1.9.0 https://github.com/geoblacklight/geoblacklight/pull/679/files#diff-fc647293c42d73c5d236eadc8a17f61cR54 assumes that Rails ujs is being used. For backwards compatibility we might need to do something similar to here: https://github.com/projectblacklight/spotlight/blob/85daf4ea3d18933cdb12cf476de221b9c6799adb/app/assets/javascripts/spotlight/pages.js.erb#L5

mejackreed:
I can also submit this to `master` as well if needed. Fixes #712


eliotjordan:
@mejackreed Seems like it would be a good idea to submit this to master as well.

PR #718, opened 2018-12-21
support Blacklight ~> 7.0

mejackreed:


mejackreed:
Yep.. I think we are on the limb of 2.0 so breaking is ok here?

Hopefully future adopters will like the catalog controller config pattern.

Issue #680, opened 2018-07-31
Reevaluate testing and integration for JavaScript


jrgriffiniii:
[Teaspoon](https://rubygems.org/gems/teaspoon/) is currently the Gem which is used for testing JavaScript, and this appears to have had waning maintenance: https://github.com/jejacks0n/teaspoon/pull/513

~Should a path towards integrating [Jasmine](https://jasmine.github.io/) be explored?  Rails 5.1 releases have used [Webpacker](https://github.com/rails/webpacker) (which would provide support for ES6 and [Jest](https://jestjs.io/)).  Perhaps these might be possible as well.~

Webpack is now supported using the `webpacker` Gem as of the merging of https://github.com/geoblacklight/geoblacklight/pull/729. However, Teaspoon now *must* be removed as a priority item given that it blocks support for Rails 6.y.z releases (please see the RubyGems page for the Gem linked above).

jrgriffiniii:
After discussing this with @KarenMajewicz @ewlarson @hudajkhan @christinach and @eliotjordan , it was decided that attempting to support [Webpacker](https://github.com/rails/webpacker) in a future 2.x release would be a reasonable first step.

jrgriffiniii:
https://github.com/jrgriffiniii/geoblacklight/tree/issues-680-jrgriffiniii-webpacker addresses the initial steps towards Webpacker integration, but I can see about migrating some simple JS functionality for feature testing with this new approach.

jrgriffiniii:
74e118e worked for me locally, but I'm still working to debug this on CircleCI: https://circleci.com/gh/jrgriffiniii/geoblacklight/135

jrgriffiniii:
783d2a6 fixed a problem I couldn't otherwise work-around with using Webpack to compile the manifest before testing on CI.  I'll need to squash my commits and review the documentation before issuing a PR.

jrgriffiniii:
This should be shifted in priority given that it will block #862

jrgriffiniii:
There are several JavaScript testing frameworks worthy of consideration for this. Two popular solutions I could be interested in seeing explored are:

- [Mocha](https://mochajs.org/)/[Chai](https://www.chaijs.com/)
- [Jest](https://jestjs.io/)/[Chai](https://www.chaijs.com/)
- [Testing Library](https://testing-library.com/docs/dom-testing-library/intro)

[...]

PR #729, opened 2019-01-23
Integrates Webpacker for managing JavaScript assets

jrgriffiniii:
Addressing #680, this integrates Webpacker for managing JavaScript assets along with:
- Configuring the ES6 transpilation to default to Babel
- Introduces Foreman to run `webpack-dev-server` in parallel with the Rails server
- Ensuring that the `application.js` pack is included by default in GeoBlacklight
implementations


mejackreed:
Does this necessitate a GeoBlacklight adopter to use Webpacker now?

jrgriffiniii:
It would require an adopter to run `bundle exec rails webpacker:compile` once before `rails server` or running the server in production in order to generate the `manifest.js`.

One alternative could be to provide a README section for including the `application.js` pack in the overridden `app/views/layouts/blacklight/base.html.erb`, remove that override from this PR, and then even this shouldn't be necessary.

jrgriffiniii:
This definitely doesn't leave anyone in the position where they need to abandon sprockets at all.  But I can certainly improve this in order to ensure that Webpacker usage is completely optional for running Rake tasks and deployment.

mejackreed:
Thats great to hear. No objection from me, just wanted to clarify it all so I could better understand. We are still dipping our toes into the webpacker land. Your experience here is much appreciated!

app/views/layouts/blacklight/base.html.erb
```diff
@@ -13,7 +13,7 @@
     <%= favicon_link_tag %>
     <%= stylesheet_link_tag "application", media: "all" %>
     <%= javascript_include_tag "application" %>
-    <%= javascript_pack_tag 'application' %>
+    <%#= Uncomment this to load the application pack javascript_pack_tag 'application' %>
```
eliotjordan: Since this partial is embedded in the gem, maybe we should just remove this file and provide instructions on how to add this in their local instances.

eliotjordan:
FWIW, I'm not opposed to just requiring webpacker, but I can see how the slow approach is beneficial. It will be useful to have webpacker already in place when we take steps towards overhauling the GBL javascript. Maybe version 3.0?

mejackreed:
For sure. FWIW my concern about requiring webpacker is making the GeoBlacklight 1.x -> 2.0 pathway more difficult without a clear deprecation notice in 1.x.

Also making sure that GeoBlacklight installations that are highly javascript customized can integrate with the webpacker approach.

jrgriffiniii:
I'm working to get this restructured so that a generator can ease the process of integrating Webpacker after an adopter has upgraded to release 2.0.  I'm also working to complete and improve documentation [on the Wiki](https://github.com/geoblacklight/geoblacklight/wiki/GeoBlacklight-2.0-Upgrade-Guide#webpacker)

jrgriffiniii:
`geoblacklight:webpacker` is now a generator which can be run in order to create the necessary Webpack directories and files within a GeoBlacklight app.  These aren't used by default unless the `javascript_pack_tag` helper is invoked in a view template (and even if it is, there is no default JavaScript which would introduce a conflict with the existing JavaScript offered by Sprockets).

New installations will have have this installed, but again, it shouldn't raise any issues unless Yarn or NPM is not available on the server environment where the *new* GeoBlacklight app. is being installed, or, where the GeoBlacklight app. has `rails generate geoblacklight:webpacker` explicitly run.

As for actually working with Webpacker for JavaScript and SCSS, I think that there is very little which can be offered for a 2.0.0 release.  I suspect that there are backwards-compatible changes which could be introduced by beginning the process of porting existing JS into ES9(10?), but I don't see the benefit in delaying any early 2.0.x releases while adopters identify more urgent issues.  I agree with the proposal to work to make Webpacker mandatory by a 3.0.0 release (along with other JavaScript standards, like linting, testing, perhaps TypeScript...)

eliotjordan:
I like the generator approach. 👍

eliotjordan:
@mejackreed Any further thoughts? I'm inclined to merge.

mejackreed:
I'm good w/ :shipit: . Just was curious about those upgrade paths and scenarios. Nice job @jrgriffiniii !

PR #732, opened 2019-01-23
Ensures that JSON representations of Solr Documents can be retrieved

### LINKED ISSUE #711: Requesting a document in JSON format does not return solr document
http://localhost:3000/catalog/tufts-cambridgegrid100-04.json

Returns this instead of solr doc:

```json
{
  "links": {
    "self": "http://localhost:3000/catalog/tufts-cambridgegrid100-04"
     },
  "data": {
    "id": "tufts-cambridgegrid100-04",
    "attributes": {
    "dc_description_s": "<div class=\"truncate-abstract\">This polygon dataset is a 100 foot grid overlay for Cambridge, MA.</div>",
    "dc_publisher_s": "<span itemprop=\"publisher\">Cambridge (Mass.) Geographic Information Systems</span>",
    "dct_temporal_sm": "<span itemprop=\"temporal\">2004</span>",
    "dct_provenance_s": "<a href=\"/catalog?f%5Bdct_provenance_s%5D%5B%5D=Tufts\">Tufts</a>"
    }
  }
}
```

jrgriffiniii:
Resolves #711, but I believe that the update to `CatalogController` will likely need to be added manually for existing adopters.  I can add this to the Wiki following https://github.com/geoblacklight/geoblacklight/wiki/GeoBlacklight-1.0-Upgrade-Guide as an example.

### DIFF
--- a/lib/generators/geoblacklight/templates/catalog_controller.rb
+++ b/lib/generators/geoblacklight/templates/catalog_controller.rb
@@ -6,6 +6,12 @@ class CatalogController < ApplicationController
   include Blacklight::Catalog
 
   configure_blacklight do |config|
+
+    # Ensures that JSON representations of Solr Documents can be retrieved using
+    # the path /catalog/:id/raw
+    # Please see https://github.com/projectblacklight/blacklight/pull/2006/
+    config.raw_endpoint.enabled = true
+
     ## Default parameters to send to solr for all search-like requests. See also SolrHelper#solr_search_params
     ## @see https://lucene.apache.org/solr/guide/6_6/common-query-parameters.html
     ## @see https://lucene.apache.org/solr/guide/6_6/the-dismax-query-parser.html#TheDisMaxQueryParser-Theq.altParameter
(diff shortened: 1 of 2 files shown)


jrgriffiniii:
I'm sorry for missing this comment earlier, but this actually uses a separate route: `/raw` (i. e. `catalog/:id/raw`): https://github.com/geoblacklight/geoblacklight/pull/732/files#diff-66e8d5275f70f4e66a363591449594e7R69

This is supported by Blacklight 7.x core, so it requires minimum configuration.  Does this still raise issues regarding routing paths?

eliotjordan:
@jrgriffiniii Ah! I just glossed over the issue in blacklight. I should have read a little deeper. No problem with this approach on my end.

eliotjordan:
We just need to remember to add the raw endpoint line to the upgrade notes.

```
config.raw_endpoint.enabled = true
```

jrgriffiniii:
I've added this to the Wiki: https://github.com/geoblacklight/geoblacklight/wiki/GeoBlacklight-2.0-Upgrade-Guide

eliotjordan:
@jrgriffiniii The circle checks are not running for some reason. Maybe give them a nudge/rebase?

Issue #606, opened 2018-01-30
Index map feature idea: highlight selected feature


kgjenkins:
### Descriptive summary

When a user clicks a feature, highlight the feature on the map (with a different color, for example) to help the user confirm which feature's data is being displayed

mejackreed:
👍 great suggestion

PR #759, opened 2019-01-29
Index map enhancements

kevinkidwell:
This PR covers the attached issue (#606) to add a selected color to the index map feature when clicked. We did a bit of cleanup and added all of the styling in the `settings.yml` file to be able to edit colors/styles in a single location and then pull them into the javascript. We also defaulted the index map feature color to the available green for all maps rather than having point index maps be blue by default. Blue is now the selected color to match with what is being used for the download buttons.

We also added in a download button in the Downloads section for the selected feature's download link so that all downloads can be accessed from that sidebar area. This button only currently applies to documents where `document_downloadable` is true.

In the feature viewer info area, we adjusted the styling a bit to match what is being done with the metadata for index maps. This was just to add some structure to that section and better definition between headings and content of the feature info.


cappadona:
@eliotjordan I'm unable to consistently reproduce the failure when running the test suite locally against rails `5.1.6`. It tripped the first time I ran the suite, but not again since. Anyone see something we've missed?

eliotjordan:
@cappadona We've definitely seen this before in other feature tests, although it's been a while since it popped up. I reran the test suite, and it passed. 🤷‍♂️ We might be able to adjust some the capybara parameters if this happens again. It's hard to fix, if you can't reproduce the problem.

cappadona:
Thanks! 👍

app/assets/javascripts/geoblacklight/templates/index_map_info.hbs
```diff
@@ -12,26 +14,26 @@
         <img src="{{thumbnailUrl}}">
       {{/if}}
     {{/if}}
-    <dl>
+    <dl class="row dl-invert document-metadata">
       {{#if websiteUrl}}
-        <dt>Website</dt>
-        <dd><a href="{{websiteUrl}}">{{websiteUrl}}</a></dd>
+        <dt class="blacklight-dc_creator_sm col-sm-3">Website:</dt>
```
mejackreed: Could we use a more generic class here? This seems to be a specific class focused on `dc_creator_sm`

kevinkidwell:
@eliotjordan @mejackreed I just pushed up a change for that. I removed that class since all of the styling was being inherited elsewhere and adjusted the CSS selector for the margin so it wasn't using the class.

eliotjordan:
@kevinkidwell Well... that didn't work. 😞

eliotjordan:
@kevinkidwell @cappadona One more thing to try. Remove that extra click and swap these two lines:

```
expect(page).to have_css 'h3', text: 'Dabao Kinbōzu -- ダバオ近傍圖'
expect(page).to have_css 'a img[src="https://stacks.stanford.edu/image/iiif/zh828kt2136%2Fzh828kt2136_00_0001/full/!400,400/0/default.jpg"]'
```

I tried this in my clone of your branch and it seems like it might work.

kevinkidwell:
@eliotjordan It looks like that did it! Thanks for working on that and figuring it out!

PR #763, opened 2019-01-31
Add field to solr document that indicates if a record should be suppressed

### LINKED ISSUE #737: Add suppressed field to solr document
- Add a suppressed field to the Solr document to indicate if a record should be suppressed in the search results. 

- Add a search builder to hide these records.

This is useful, for example, in atlases where individual map sheets share nearly identical metadata. Hundreds of nearly identical maps make searching much less useful. Princeton's Geoblacklight instance currently does this for it's collection of Sanborn maps.

Example search builder code:

https://github.com/pulibrary/pulmap/blob/68e31988b73a4230b7aed630735b4af58cb683a9/app/models/search_builder.rb#L10-L18

eliotjordan:
Adds the `suppressed_b` field to the solr doc to indicate if a record should be suppressed in the search.

Closes #737

### DIFF
--- a/app/models/concerns/geoblacklight/spatial_search_behavior.rb
+++ b/app/models/concerns/geoblacklight/spatial_search_behavior.rb
@@ -3,7 +3,7 @@ module Geoblacklight
     extend ActiveSupport::Concern
 
     included do
-      self.default_processor_chain += [:add_spatial_params]
+      self.default_processor_chain += [:add_spatial_params, :hide_suppressed_records]
     end
 
     ##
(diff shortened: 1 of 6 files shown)


mejackreed:
Is there an expectation that a user gets to a show page for these in a different way?

eliotjordan:
@mejackreed The main use case here is suppressing child records in a parent-child relationship like sheets in a Sanborn an atlas. The parent record is searchable, and the child records are available through the `relations` panel.

This line allows them to show up in the search results when clicking on the "Browse all records" records link:

https://github.com/geoblacklight/geoblacklight/blob/65189692342b47fa8825df099586bd2e28f23c98/app/models/concerns/geoblacklight/spatial_search_behavior.rb#L53

Example of a Sanborn map where child records are suppressed:
https://maps.princeton.edu/catalog/princeton-dz010s95f

Suppressed records from that atlas appearing in a related record search:
https://maps.princeton.edu/catalog?f%5Bdct_source_sm%5D%5B%5D=princeton-dz010s95f

mejackreed:
Gotcha thanks, I think I understand now. This answers my questions. 👍 

One thing I noticed is an issue with the URl encoding on the search from the Princeton site, but thats another issue :)

eliotjordan:
Yeah. I just realized that there is an outstanding bug report for that. Hopefully we can fix that soon. 🤞

PR #765, opened 2019-01-31
Issue 324 bboxrelevancy

ewlarson:
This PR adds optional spatial search overlapRatio relevancy support. Configurable via the presence of Settings.OVERLAP_RATIO_BOOST.

When a bbox param is present, the Settings.OVERLAP_RATIO_BOOST value will boost results based on the overlap ratio of the result bboxes.

Includes a modification to our Solr schema to include a _bboxtype fieldtype. Requires Solr 7.4+ 

Thanks to:
@hudajkhan / Feature development
@ewlarson / Rspec tests

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<schema name="geoblacklight-schema" version="1.6">
+<schema name="geoblacklight-schema" version="1.7">
   <uniqueKey>layer_slug_s</uniqueKey>
   <fields>
     <field name="_version_" type="long"   stored="true" indexed="true"/>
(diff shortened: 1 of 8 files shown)


solr/conf/schema.xml
```diff
@@ -187,4 +194,7 @@
   <copyField source="dct_provenance_s" dest="suggest"/>
   <copyField source="dc_subject_sm" dest="suggest"/>
   <copyField source="dct_spatial_sm" dest="suggest"/>
-</schema>
+  
+  <!-- for bbox value -->
+  <copyField source="*_geom" dest="*_bboxtype"/>
```
mejackreed: Can we assume that all `*_geom` fields can be copied into `_*bboxtype`? If I provide point data here, what happens?

hudajkhan: I will try to test that out since that is something we should evaluate.  If this copying over doesn't work for point fields, don't know if there is such a thing as a conditional copy possible in schema (i.e. If solr_geom is bbox-compatible, copy over)

eliotjordan: Out current schema uses the "ENVELOPE WKT" syntax in the `solr_geom` field which will copy over fine. Instead of copying all `_geom` fields, how about just `solr_geom`?

ewlarson: @hudajkhan @mejackreed @eliotjordan We are all on the same page here. Let's limit the copy to just the solr_geom field. Want to push that commit @hudajkhan?

hudajkhan: Will make this change now.

Issue #344, opened 2015-07-24
Add "Open in ArcGIS.com" button


mejackreed:
This feature has already been investigated in conjunction with #341 . 

There seem to be some problems though associated with adding this feature which seem to detract from use experience.
- ArcGIS.com seems to take url parameters to add a Esri webservice or wms layer. This "feature" is undocumented from Esri and its unclear how long it will even work
  - `url` parameter for an Esri webservice `http://www.arcgis.com/home/webmap/viewer.html?url=http://sampleserver6.arcgisonline.com/arcgis/rest/services/Census/MapServer/`
  - `wms` parameter for a wms endpoint. Note this takes a `visibleLayers` parameter which indicates what layer to show. `http://www.arcgis.com/home/webmap/viewer.html?wms=https%3A%2F%2Fgeowebservices.stanford.edu%2Fgeoserver%2Fwms&visibleLayers=druid%3Acz128vq0535` 
    - **Note** this doesn't work for all WMS services in fixture data
    - Once in ArcGIS.com as a WMS service, there is not really any way for that data to be used or styled
  - ArcGIS.com doesn't seem to provide for a way to import data directly into a mapview. CartoDB oneclick offers this in a really nice way

mejackreed:
My thought on this is punt on it for now, until someone actually needs this in the future.

ajturner:
This is what we used for adding open in ArcGIS to data.gov

mejackreed:
:+1: ah good to know... thanks @ajturner

jgravois:
based on our discussions at Geo4LibCamp, i whipped up a quick example to show how y'all could use `@esri/arcgis-rest-request` to help ArcGIS users login and create an item of their own (pointing at WFS services) and then launch a new map viewer session.

**please** holler if you have any trouble implementing something similar here.

https://github.com/jgravois/geoblacklight-arcgis

![black_to_arc](https://user-images.githubusercontent.com/3011734/52450515-2e5ee780-2af0-11e9-8e42-cc2fc575951b.gif)

PR #781, opened 2019-02-08
Create an "Open in ArcGIS" button for Esri web services

mejackreed:
Part of #344 

This is an initial take on tackling #344 and only implements an "Open in ArcGIS Online" button for content that is already coming from an Esri web service.

This work comes out of discussions at Geo4LibCamp 2019. Thanks so much for help on this from @eliotjordan, @karenmajewicz, @ajturner, @jgravois, and @dmfenton . A more complete implementation will come with details provided in #344, but we wanted to get the initial pieces in front of users for feedback.

openinarcgis


app/helpers/arcgis_helper.rb
```diff
@@ -0,0 +1,12 @@
+module ArcgisHelper
+  ##
+  # Generates an ArcGIS.com viewer url with params that can open content
+  # directly
+  # See: https://doc.arcgis.com/en/arcgis-online/reference/use-url-parameters.htm
+  def arcgis_link(arcgis_url)
```
eliotjordan: For consistency, it would probably be good pluralize the param name here: `arcgis_urls`.

andrewbattista:
This looks so great! Way to go, @mejackreed @eliotjordan John, and others!

PR #790, opened 2019-05-11
Modifies the CircleCI configuration to use Docker containers for Solr

jrgriffiniii:
This should ensure that failures to shut down `solr_wrapper` don't break builds for CircleCI (please see https://circleci.com/gh/geoblacklight/geoblacklight/888 for reference).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,8 @@ jobs:
           CI: true
           RAILS_ENV: test
           RAILS_VERSION: 5.2.0
+      - image: solr:7-alpine
+        command: bin/solr -cloud -noprompt -f -p 8983
     steps:
       - checkout
       # Update chrome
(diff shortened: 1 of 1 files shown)


mejackreed:
Thanks for doing this @jrgriffiniii ! One thing now that I'm thinking about is that this provides a different method of running the CI build now on Circle vs. local. And the Circle build steps seem to get more complicated. 

How are y'all handling this at princeton for other applications? Are you using the dockerized Solr for everything? (not a blocking review, just thinking about how we maintain this long term).

jrgriffiniii:
You're welcome, thank you for maintaining the build (and sorry for not getting this better documented).  

At Princeton the usage of Dockerized Solr varies, but we currently do take this approach with Figgy (https://github.com/pulibrary/figgy/blob/master/.circleci/config.yml#L73), and the community builds for Samvera Components have been following this pattern also (please see https://github.com/samvera-labs/samvera-circleci-orb/blob/master/src/executors/ruby_fcrepo_solr.yml#L15).

After becoming a bit familiar with CircleCI Orbs (https://circleci.com/orbs/), I would be eager to try and take an approach for GeoBlacklight where the CircleCI configuration is managed within a separate GitHub repository, but I felt as if this would only further block any solutions I could immediately provide for the build at this time.

PR #808, opened 2019-06-20
Fixes FGDC references documentation

eliotjordan:


schema/references.md
```diff
@@ -11,7 +11,7 @@ Direct download file | `http://schema.org/downloadUrl` | Direct file download fe
 Data dictionary / documentation download | `http://lccn.loc.gov/sh85035852` | Direct documentation download link|
 Full layer description | `http://schema.org/url` | Further descriptive information about layer|
 Metadata in ISO 19139 | `http://www.isotc211.org/schemas/2005/gmd/` | Structured metadata in ISO standard expressed as XML|
```
mejackreed: Does this one also need modification?

eliotjordan: ISO has the slash. 🤷‍♂

eliotjordan: https://github.com/geoblacklight/geoblacklight/blob/f9f1cd4b13f0cc1344230d8d3b7efb2a489714f2/lib/geoblacklight/constants.rb#L11

mejackreed: yikes!

Maybe we could do an update to support either case too? (not a blocker for this PR just brainstormin')

eliotjordan: https://github.com/geoblacklight/geoblacklight/issues/809

PR #820, opened 2019-07-29
JS: update Readmore, change call syntax

### LINKED ISSUE #800: jQuery 3+ readmore-js
### Descriptive summary

GBL 2.1+ / jQuery 3.4+ 

The readmore-js library included in vendor/assets is a jquery plugin, which requires jQuery < 2.0. New versions of readmore-js do not support jQuery, with the most recent versions of the readmore-js tool having gone to vanilla javascript.
https://github.com/jedfoster/Readmore.js/issues/177

So! If you update the version of jQuery used by GBL from 1.X to 3.X, the readmore feature will break.

1. In application.js change require jquery to require jquery3
1. Load http://localhost:3000/catalog/2eddde2f-c222-41ca-bd07-2fd74a21f4de
1. See readme fail

### Fix will be...

We'll want to drop the jquery plugin version of readmore and ship the most recent 3.0 version (beta-1 at this point). Then in modules/item.js we'll change the call from jquery plugin syntax to vanilla js.

From: 
`
  $('.truncate-abstract').readmore({
    maxHeight: 60
  });
`

To:
`
  new Readmore(
    '.truncate-abstract',
    {
      maxHeight: 60
    }
  );
`

### Related work

B1G Geoportal: https://github.com/BTAA-Geospatial-Data-Project/geoportal/issues/224

ewlarson:
Update Readmore-js to v3.0.0-beta-1 (latest release). Readmore v3 is no longer a jQuery plugin, so the GBL item module needs editing to call the feature via: new Readmore(...). Add simple js spec to test Readmore is defined.

Fixes #800


app/assets/javascripts/geoblacklight/modules/item.js
```diff
@@ -9,7 +9,5 @@ Blacklight.onLoad(function() {
     viewer = new window['GeoBlacklight']['Viewer'][viewerName](element);
   });
 
-  $('.truncate-abstract').readmore({
```
mejackreed: Is this technically a breaking change? Is it possible to check which version is around and do the right thing based off that? (with a note to remove for GeoBlacklight 3.0?)

mejackreed: Maybe its not a breaking change.. the interface I guess stays the same

ewlarson: Yeah, no breaking change here. Just leaving the jQuery plugin version for a vanilla JS implementation. View markup doesn't change at all.

lib/generators/geoblacklight/templates/geoblacklight.js
```diff
@@ -7,4 +7,4 @@
 //= require geoblacklight/downloaders
 //= require leaflet-iiif
 //= require esri-leaflet
-//= require readmore.min
+//= require readmore
```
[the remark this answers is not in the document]
mejackreed: Oh good call. Thanks for raising this @dunn

PR #828, opened 2019-08-01
Restructures the CircleCI config. for parallel builds

### LINKED ISSUE #789: Parallelize CircleCI builds
Currently there are only two concurrent jobs which run against Rails releases 5.2.0 and 5.1.6 using Ruby releases 2.5.1 and 2.4.4, and one following job which runs Rubocop (on 2.5.1).  Perhaps the `engine_cart` generation, Solr Collection creation, and RSpec test steps could be separated into different build steps?
This wouldn't necessarily increase the speed of the build, but one could more easily determine whether or not a branch is going to fail - caching might also be more readily managed.

jrgriffiniii:
Resolves #789 by addressing the following:

- Restructuring the CircleCI into three build jobs: bundle, rubocop, and test
- Ensuring bundler Gems and the engine_cart internal test app. is cached on CircleCI
- Ensuring that the tests are run in parallel between containers

Also handles cases where OpenSSL errors are encountered when requesting metadata files

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,91 +1,106 @@
 ---
-version: 2
+version: 2.1
 jobs:
-  rails_5_2_3:
+  bundle:
+    parameters:
+      ruby_version:
+        type: string
+      rails_version:
+        type: string
     working_directory: ~/geoblacklight
     docker:
-      - image: circleci/ruby:2.5.1-node-browsers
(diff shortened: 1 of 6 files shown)


jrgriffiniii:
This still isn't running on CircleCI.  I'll attempt to push directly to a branch on this repository.

PR #833, opened 2019-08-01
Support reference keys with or without slashes

### LINKED ISSUE #809: Support ISO reference keys with and without a slash
Support ISO metadata reference keys with and without a slash:

https://github.com/geoblacklight/geoblacklight/blob/master/lib/geoblacklight/constants.rb#L11

`http://www.isotc211.org/schemas/2005/gmd/`

and

`http://www.isotc211.org/schemas/2005/gmd`

christinach:
- Updates iso19139 key in Constants::URI not to have a trailing slash
- Removes trailing slashes from reference key if they exist

closes #809


lib/geoblacklight/reference.rb
```diff
@@ -38,7 +38,13 @@ def to_hash
     # The URI used for this instance's creation
     # @return [String]
     def uri
-      @reference[0]
+      @reference[0] == 'http://www.isotc211.org/schemas/2005/gmd/' ? iso19139_parse(@reference[0]) : @reference[0]
```
mejackreed: Could this be generalized to just remove trailing slashes from anything?

christinach: yes I can update it

eliotjordan: I agree, we can just generalize this. Otherwise, PR looks good.

mejackreed: 👍 same.. maybe just even a regex? `'http://example.com///'.sub(/\/*$/, '')`

christinach: @mejackreed  now i see your comment; yes I can update to remove any trailing slashes not just one and use a regex.

PR #836, opened 2019-08-05
Add support for oembed urls in dct_references and viewers

### LINKED ISSUE #826: Feature enahancement: Add dct_reference for oembed content
### Descriptive summary
This is a feature enhancement to add a dct_refererence field for oembed content. This could take advantage of the https://github.com/sul-dlss/blacklight-oembed gem

mejackreed:
This PR fixes #826 by enabling support for oembed content in dct_references. A few comments about it.

oembed

 - ~Users updating to the release of GeoBlacklight with this will need to modify their `catalog_controller.rb` file to include the `default_oembed` partial.~ Using "viewer" approach now. No updates needed.
 - This approach does not use the "register" oembed provider approach, but rather takes a "simpler" just give me the embed url. Perhaps this might be problematic, but it feels simpler than keeping a registry of providers or custom options for them.
 - No js test is added, as I didn't want to increase test suite overhead.. if others prefer this is added, happy to add it.
 - ~This was built with the approach that maybe you also want to show the map.. but we could also do it in place of the map.. I'm not sure. Any suggestions?~ per suggestions using single viewer approach.


mejackreed:
Going to add a [WIP] label on this so we can discuss at standup tomorrow.

PR #837, opened 2019-08-05
Configure the display of the Carto and ArcGIS links

### LINKED ISSUE #810: Add the ability to configure display of carto and ArcGIS export links
We want to display the "Open in ArcGIS" button, but not the the "Open in Carto" button. It would be nice to have this configurable instead of having to override the exports partial or the cartodb_reference method in the solr document.

christinach:
closes #810

### DIFF
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -170,7 +170,7 @@ module GeoblacklightHelper
       t('geoblacklight.tools.open_carto'),
       carto_link(@document.carto_reference),
       target: '_blank'
-    ) if @document.carto_reference.present? && Settings.CARTO_DISPLAY
+    ) if @document.carto_reference.present? && !Settings.CARTO_HIDE
   end
 
   def display_arcgis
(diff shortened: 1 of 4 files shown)


app/helpers/geoblacklight_helper.rb
```diff
@@ -164,6 +164,23 @@ def geoblacklight_basemap
     blacklight_config.basemap_provider || 'positron'
   end
 
+  def display_carto
+    return link_to(
+      content_tag(:span, '', class: 'geoblacklight geoblacklight-carto') +
+      t('geoblacklight.tools.open_carto'),
+      carto_link(@document.carto_reference),
+      target: '_blank'
+    ) if @document.carto_reference.present? && Settings.CARTO_DISPLAY
```
eliotjordan: I wonder if the default should be to display Carto and ArcGIS? Or put another way, test if CARTO_DISPLAY and ARCGIS_DISPLAY are `false` rather than true. If a user does an upgrade to GBL and does not update the settings.yml file, they will lose the links and may not know why.

christinach: @eliotjordan thank you for thinking about this and reviewing it. As we further discussed I updated the settings so that hide the arcgis and carto links are optional for the users.

PR #840, opened 2019-08-08
Inline svgs

ewlarson:
This PR moves us from the geoblacklight-icons plugin and font-based iconography to inline SVGs. The implementation is largely based on the [blacklight_icon helper](https://github.com/projectblacklight/blacklight/blob/master/app/helpers/blacklight/icon_helper_behavior.rb) developed in Blacklight, but includes an additional process to render aria-labels for each SVG.

Screen Shot 2019-08-08 at 2 42 58 PM

## Icons
First we bring all the icons from geoblacklight-icons over into GeoBlacklight within the /app/assets/images/blacklight dir. I used the [IcoMoon](https://icomoon.io/) App to migrate the SVGs and TTF glyphs into a single IcoMoon project. IcoMoon helped standardize the icon color set. Import the geoblacklight-icons.json file to IcoMoon to recreate this icon set in their application. 

Screen Shot 2019-08-09 at 11 02 51 AM

I also used [SVGOMG](https://jakearchibald.github.io/svgomg/) to strip unnecessary elements and attributes from the SVGs, and to remove hard coded height and width values.

Screen Shot 2019-08-08 at 2 08 03 PM

## Icon Use 

Our geoblacklight_icon method, which previously wrote spans with classes for our font-based icons is updated to return SVG icons or an empty HTML element.

```ruby
  ##
  # Returns an SVG icon or empty HTML span element
  # @return [SVG or HTML tag]
  def geoblacklight_icon(name)
    icon_name = name ? name.to_s.parameterize : 'none'
    if asset_exists?(icon_name)
      queue_icon_aria_label(icon_name)
      blacklight_icon(icon_name)
    else
      render_empty_span('icon-missing geoblacklight-none')
    end
  end
```

First we check that the icon indeed exists, then we queue the icon's name for aria description, and finally call blacklight_icon to fetch the SVG. If no icon is found, we return an empty span. 

## Accessibility

Each SVG in the application has an aria-labelledby value, example:

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 32" aria-labelledby="aria-label-berkeley">
  <title>
    berkeley
  </title>
...
```

When we call for an icon using the geoblacklight_icon helper, the icon names are collected and when the page is rendered, we populate a list of aria-label definitions via:

```ruby
  ##
  # Render a div of divs describing aria-labelledby values
  # @return [HTML tag]
  def render_aria_labels(aria_labels)
    return unless aria_labels
    content_tag :div, id: 'aria-labels', class: 'sr-only sr-only-focusable' do
      aria_labels.each do |label|
        concat(render_aria_label(label))
      end
    end
  end
```

This helper renders a div of divs describing the aria-labelledby values:

Screen Shot 2019-08-08 at 2 28 58 PM

The render_aria_labels(@aria_labels) method is called from app/views/shared/_header_navbar.html.erb because this is the only view file within GeoBlacklight that renders across every view the application produces.

## I18n

Finally, Rails' I18n translations are used to draft the aria values on the page.
 
```ruby
  # Render a div describing a aria-labelledby value
  # @return [HTML tag]
  def render_aria_label(label)
    content_tag :div, id: "aria-label-#{label}" do
      I18n.t("geoblacklight.aria-labels.#{label}")
    end
  end
```

## Adding Additional SVGs

Adding new SVGs to GeoBlacklight would entail:

1. Adding the new SVG icon to app/assets/images/blacklight
1. Ensuring the SVG icon had an aria-labelledby attribute with a value matching it's own filename
1. Writing the :en I18n entry for the aria label via: 

```yaml
    ...
    aria-labels:
      arrow-circle-down: Arrow within a circle, pointing down (icon)
      baruch-cuny: Baruch College (logo)
      new-example-here: New Example description text
```


app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
```diff
@@ -1,60 +1,50 @@
...
-}
-
-.geoblacklight-carto {
-  @extend .fa, .fa-map-marker;
-}
-
-// DEPRECATED: Will be removed in GeoBlacklight v2.0
-.geoblacklight-cartodb {
-  @extend .geoblacklight-cartodb;
+// SVG icons via FontAwesome and IcoMoon, see icomoon.json for project data
```
mejackreed: Which file is `icomoon.json`?

app/assets/images/blacklight/geoblacklight-icons.json
```diff
@@ -0,0 +1,2079 @@
+{
```
mejackreed: Is this auto generated somehow? 
How would we add new icons?

Is this so they all appear the same color or are there other purposes?

app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
ewlarson: Sorry! Tis the geoblacklight-icons.json file

app/assets/images/blacklight/geoblacklight-icons.json
ewlarson: This file is auto-generated from IcoMoon. It would help us add additional icons (from Font Awesome or elsewhere) and keep them all color-coordinated.

Try going to IcoMoon.io, click the IcoMoon App button, drag and drop this file onto the screen. You should see all the icons at play here.

mejackreed:
I'm seeing some odd stuff with the esri logo:

Screen Shot 2019-08-08 at 3 28 14 PM

app/assets/images/blacklight/geoblacklight-icons.json
mejackreed: Ok thanks for that 👍

app/assets/stylesheets/geoblacklight/modules/icon-customization.scss
mejackreed: 👍

ewlarson:
Strange. I'll look into that Esri logo bug.

ewlarson:
A few todos left here:

- [ ] Fix show_tools icon placement (render icon within anchor element)
- [ ] Fix Esri icon (02236876-9c21-42f6-9870-d2562da8e44f)
- [ ] Fix Derived Datasets icon (nyu_2451_34636)
- [ ] Add wiki documentation for IcoMoon, adding additional icons, changing icon color.

ewlarson:
Screenshot to show improved show_tools icon placement.

Screen Shot 2019-08-09 at 11 11 19 AM

PR #846, opened 2019-08-09
Add new reference fields and enable multiple downloads

### LINKED ISSUE #304: Should the formats facet include generated download types?
I had some feedback from users that they were looking for GeoJSON data and could not find any using the data format facet. Should the format facet include generated download types? I think this is possible...

eliotjordan:
Closes #304 

- Adds new reference fields that will replace `dct_references` in the next major version while maintaining backwards compatibility with existing documents.

  ```
  "downloads_sm": [
    "{\"type\":\"GeoTIFF\",\"url\":\"http://stacks.stanford.edu/file/druid:dp018hs9766/data.zip\",\"label\":\"Original GeoTIFF\"}"
  ],
  "metadata_sm": [
    "{\"type\":\"URL\",\"url\":\"http://purl.stanford.edu/dp018hs9766\"}",
    "{\"type\":\"MODS\",\"url\":\"http://purl.stanford.edu/dp018hs9766.mods\"}",
    "{\"type\":\"ISO19139\",\"url\":\"https://raw.githubusercontent.com/OpenGeoMetadata/edu.stanford.purl/master/dp/018/hs/9766/iso19139.xml\"}"
  ],
  "webservices_sm": [
    "{\"type\":\"WCS\",\"url\":\"https://geowebservices-restricted.stanford.edu/geoserver/wcs\"}",
    "{\"type\":\"WMS\",\"url\":\"https://geowebservices-restricted.stanford.edu/geoserver/wms\",\"layerId\":\"druid:dp018hs9766\",\"exportFormats\":[\"GeoTIFF\"]}"
  ]
  ```

- Enables multiple downloads
- Brings coverage back up to 100%


eliotjordan:
Follow up tasks:

- Review, update, and merge https://github.com/OpenGeoMetadata/GeoCombine/pull/91
- Add documentation to the wiki for using new references and add deprecation notes to the `dct_references` and `layer_id_s` fields.
- Update the JSON schema and bump the version.

lib/geoblacklight/item_viewer.rb
```diff
@@ -14,41 +14,43 @@ def viewer_endpoint
...
 
-    def wms
-      @references.wms
-    end
+    private
 
-    def iiif
-      @references.iiif
-    end
+      def wms
```
mejackreed: Is this a rubocop thing? I'd be fine to ignore if it makes things simpler for you.

eliotjordan: Not a rubocop cop thing, just a personal preference/reflex. We tend to do this in Princeton projects as well. I can adjust this to match the rest of the codebase.

mejackreed: Gotcha. I'd be fine for this maybe in another PR.. but adding the style things here makes the review a bit more difficult.

eliotjordan: Actually, rubocop is complaining about this. The `Layout/IndentationWidth` cop which seems a bit general to turn off for the whole file. We have to use `Layout/IndentationConsistency` and set `EnforcedStyle: indented_internal_methods`

eliotjordan: Or not... yeah, we'll just have to exclude the file from the `IndentationConsistency` cop.

eliotjordan: Fixed.

eliotjordan:
@mejackreed Certainly. PRs welcome. I figured there would be some back and forth as it this essentially a proof-of-concept and is a little rough around the edges.

@kgjenkins Any thoughts on this? I know you've been looking for a solution to this problem for awhile now!

eliotjordan:
Removed some extra fixture documents that were part of code that moved to GeoCombine.

PR #850, opened 2019-08-09
B1G Fixture Updates

ewlarson:
Fixes two JS/CORS issues related to B1G contributed fixtures. Switches to a new esri-dynamic-layer-single-layer fixture, which U of MN controls to avoid future CORS issues.


ewlarson:
Closing this PR. The original fixture is working fine. We can revisit this if the problem returns.

PR #865, opened 2020-01-21
Bug fix: Exportable actions should return 404 if doc not found

### LINKED ISSUE #859: ActionView::Template::Error - Bad Document Identifier - GBL Exportable Routes
### Descriptive summary

GeoBlacklight 2.2+ (probably older versions too).

Have seen this issue within a couple different GBL instances running in production (B1G Geoportal and GeoData@WI). Once your instance is crawled by bots, they tend to come back around periodically. If you remove a document from your index, the bots still return to try and crawl it. That's fine if they hit the CatalogController#show action, because that will 404, example:
https://geo.btaa.org/catalog/e1ec54e6-8bb4-496a-93f9-43ac901bea74

However, if the bots also try to crawl an "exportable" route, say /metadata or /web_services, that page will 500, example:
https://geo.btaa.org/catalog/e1ec54e6-8bb4-496a-93f9-43ac901bea74/metadata

It 500s because the request gets all the way to view rendering and fails calling document.references -- there is no document, so:

Screen Shot 2020-01-09 at 3 26 10 PM

### Expected behavior

I would expect GBL to return a 404, because no matching document id is to be found in Solr.

### Actual behavior

We get a 500.

1. Clone GBL master; spin up the app
1. Visit a URL for a bad document identifier: http://localhost:3000/catalog/bad-document-id-here/metadata
1. See the 500. 

### Related work

INTERESTINGLY, exportable methods we inherit from Blacklight do render. They just return a blank page, which is probably also the "wrong thing to do".

http://localhost:3000/catalog/bad-document-id-here/citation

Screen Shot 2020-01-09 at 3 30 17 PM

So! Blacklight might also want to return a code instead of rendering empty pages.

ewlarson:
Fixes #859

Opening this PR for review. This change overrides the stock BL CatalogController "action_documents" method, allowing us to raise record not found errors for exportable methods, like: /metadata, /web_services, or /citation. Currently GBL 500s on bad-document-id/metadata if no document is found, and awkwardly renders 200 for bad-document-id/citation. View issue for full details on the context for this bug.

This change _could_ also be submitted to Blacklight. I vacillate on whether GBL or BL is the best path forward here, as we'd need them to cut a release and GBL might not be ready for the latest BL release. 

Opinions requested!

### DIFF
--- a/lib/generators/geoblacklight/templates/catalog_controller.rb
+++ b/lib/generators/geoblacklight/templates/catalog_controller.rb
@@ -265,6 +265,12 @@ class CatalogController < ApplicationController
     config.autocomplete_path = 'suggest'
   end
 
+  # @return [Array] first value is a Blacklight::Solr::Response and the second
+  #                 is a list of documents
+  def action_documents
+    deprecated_response, @documents = search_service.fetch(Array(params[:id]))
+    raise Blacklight::Exceptions::RecordNotFound unless @documents.present?
 
-
+    return deprecated_response, @documents
+  end
 end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
   rails_5_2_3:
     working_directory: ~/geoblacklight
     docker:
-      - image: circleci/ruby:2.5.1-node-browsers
+      - image: circleci/ruby:2.5.7-node-browsers
         environment:
           BUNDLE_JOBS: 3
           BUNDLE_RETRY: 3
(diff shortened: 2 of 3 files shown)


mejackreed:
I'd be more inclined to getting this fixed upstream in Blacklight. I could help facilitate a release if we need to

lib/generators/geoblacklight/templates/catalog_controller.rb
```diff
@@ -265,6 +265,12 @@ class CatalogController < ApplicationController
     config.autocomplete_path = 'suggest'
   end
 
+  # @return [Array] first value is a Blacklight::Solr::Response and the second
+  #                 is a list of documents
+  def action_documents
```
mejackreed: If we do end up doing something in GeoBlacklight can we use the concern approach rather than generating this into adopters application?

ewlarson: Thanks for the feedback! I'll move this fix over to Blacklight.

ewlarson:
Closing. Opened PR upstream in Blacklight:
https://github.com/projectblacklight/blacklight/pull/2233

PR #871, opened 2020-01-24
Make suggested metadata elements optional

### LINKED ISSUE #855: Consider changing some metadata elements from required to optional
### Descriptive summary

The GeoBlacklight schema requires several elements that effectively restrict what kind of layers can be loaded into a GeoBlacklight application. Examples of records that someone might want to add to GeoBlacklight without adding a value for a required element include:

- Scanned map without bounding box coordinates (no solr_geom)
- Collection level record describing a group or subject (potentially no layer_geom_type_s, solr_geom, or provenance) 
- Collective record held by multiple institutions (multiple provenances)
- Resource held by external organization (no institutional or obvious provenance)

---------------------------------------

Additionally, there is a slight discrepancy between the required elements in the schema documentation as opposed to what is actually necessary to load into the application. This table shows the affected elements (at least as configured for the BTAA geoportal - please update if these are incorrect for default GeoBlacklight):

| element name | required by schema documentation | required for loading into application | tied to icon  display |
| -------------- | ------------------------------------ | ----------------------------------- | ------ |
dc_identifier_s | X | X |
dc_rights_s | X | X | X |
dc_title_s | X | X |
dct_provenance_s | X | X | X |
geoblacklight_version | X | |
layer_slug_s | X | X |
solr_geom | X | ? |
layer_geom_type_s | | X | X |

### Proposed Changes
This ticket proposes that we consider releasing the requirement for the following elements to facilitate more flexible records:
- [x] dct_provenance_s
- [x] solr_geom
- [x] layer_geom_type_s

Perhaps we could add another level to the schema beyond **Required** & **Optional** to indicate that these values are "**Strongly Recommended**"?

### Related work

Link to related tickets or prior related work here.

#628 (geometry type is still causing browser error for BTAA; anyone else?)

ewlarson:
The changes within this PR ensure fields solr_geom, layer_geom_type_s, and dct_provenance_s are optional and are no longer required.

See the corresponding issue for full details and documentation. Once merged, I'll open an issue to update our schema documentation.

Fixes #855


app/models/concerns/geoblacklight/solr_document.rb
```diff
@@ -45,7 +45,11 @@ def oembed
     end
 
     def same_institution?
-      fetch(Settings.FIELDS.PROVENANCE).casecmp(Settings.INSTITUTION.downcase).zero?
+      if self.has_field?(Settings.FIELDS.PROVENANCE)
+        fetch(Settings.FIELDS.PROVENANCE).casecmp(Settings.INSTITUTION.downcase).zero?
```
mejackreed: I think `fetch` allows you to provide a default value as an argument which could help remove the double conditional

ewlarson: Okay, this _could_ be rewritten as such:

```ruby
    def same_institution?
      fetch(
        Settings.FIELDS.PROVENANCE,
        self.has_field?(Settings.FIELDS.PROVENANCE)
      )
      .to_s
      .casecmp(
        Settings.INSTITUTION.downcase
      ).zero?
    end
```

But, I honestly prefer the if/else statement because it is explicit. Chaining the .to_s to fetch (to support fetch returning false) to pass along to .casecmp feels a bit mumbo jumbo.

@mejackreed let me know which way you prefer and we'll go that route.

mejackreed: What about just this?

```diff
def same_institution?
-  fetch(Settings.FIELDS.PROVENANCE).casecmp(Settings.INSTITUTION.downcase).zero?
+  fetch(Settings.FIELDS.PROVENANCE, '').casecmp(Settings.INSTITUTION.downcase).zero?
end
```

ewlarson: That's a winner. 😄

jrgriffiniii:
Until this is rebased, the tests are going be pending, but they are passing.

Issue #866, opened 2020-01-22
Error raised when running the engine cart


christinach:
after bundle install
and bundle exec rake geoblacklight:server I get the error:
```
Bundler could not find compatible versions for gem "sprockets":
  In snapshot (Gemfile.lock):
    sprockets (= 4.0.0)

  In Gemfile:
    geoblacklight was resolved to 2.2.1, which depends on
      handlebars_assets was resolved to 0.23.7, which depends on
        sprockets (>= 2.0.0)

    sass-rails (~> 5.0) was resolved to 5.1.0, which depends on
      sprockets (>= 2.8, < 4.0)

    engine_cart (~> 2.0) was resolved to 2.3.0, which depends on
      rails was resolved to 5.2.4.1, which depends on
        sprockets-rails (>= 2.0.0) was resolved to 3.2.1, which depends on
          sprockets (>= 3.0.0)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

christinach:
I was using `ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]`

jrgriffiniii:
I can confirm that I've reproduced this after invoking `bundle exec rake geoblacklight:server` when using Ruby release 2.5.7. It should be noted that this error was *not* raised when I used Ruby release `2.4.4`.

jrgriffiniii:
Rails 6.0.1 did fix this with https://github.com/rails/rails/issues/36954

mejackreed:
We have fixed this in other Blacklight gems like this: https://github.com/projectblacklight/blacklight-gallery/blob/master/spec/test_app_templates/Gemfile.extra

jrgriffiniii:
https://circleci.com/workflow-run/6396beca-04f0-4893-b924-67a17286d997 gets this resolved for Rails 5.1.z releases and Ruby version 2.5.z and 2.6.z. Builds for Ruby 2.7.0 are failing, but this appears to be a Blacklight upstream issue: https://circleci.com/gh/jrgriffiniii/geoblacklight/504

jrgriffiniii:
https://github.com/projectblacklight/blacklight/pull/2226 resolved this for Ruby 2.7.0, we may just need to wait for the next patch or minor release for Blacklight.

jrgriffiniii:
As this is no longer breaking the builds on CircleCI, I'm going to close this until it is reproduced.

PR #876, opened 2020-01-27
Updating the structure of the CircleCI config. to test against Ruby 2.6.z and 2.5.z releases

jrgriffiniii:
Advances #866 and #789 

- Updating the structure of the CircleCI config. to test against Ruby 2.6.z and 2.5.z releases along with Rails 5.1.z and 5.2.z releases
- Ensuring that sass-rails 6.0.z releases are used when generating Rails 5.1.z and 5.2.z apps
- Removing support for Rails 4.y.z releases from the Gemfile

### DIFF
--- a/lib/generators/geoblacklight/install_generator.rb
+++ b/lib/generators/geoblacklight/install_generator.rb
@@ -92,6 +92,11 @@ module Geoblacklight
     end
 
     def bundle_install
+      # Override the sass-rails problem
+      if Rails.version =~ /5\.[12]/
+        gsub_file('Gemfile', "gem 'sass-rails', '~> 5.0'", "gem 'sass-rails', '~> 6.0'")
+      end
+
       Bundler.with_clean_env do
         run 'bundle install'
       end
--- a/lib/generators/geoblacklight/install_generator.rb
+++ b/lib/generators/geoblacklight/install_generator.rb
@@ -97,11 +97,6 @@ module Geoblacklight
     end
 
     def bundle_install
-      # Override the sass-rails problem
-      if Rails.version =~ /5\.[12]/
-        gsub_file('Gemfile', "gem 'sass-rails', '~> 5.0'", "gem 'sass-rails', '~> 6.0'")
-      end
-
       Bundler.with_clean_env do
         run 'bundle install'
       end
--- a/lib/generators/geoblacklight/install_generator.rb
+++ b/lib/generators/geoblacklight/install_generator.rb
@@ -91,6 +91,11 @@ module Geoblacklight
       gsub_file('config/locales/blacklight.en.yml', 'Blacklight', 'GeoBlacklight')
     end
 
+    # Ensure that assets/images exists
+    def create_image_assets_directory
+      FileUtils.mkdir_p('app/assets/images') unless File.directory?('app/assets/images')
+    end
+
     def bundle_install
       # Override the sass-rails problem
       if Rails.version =~ /5\.[12]/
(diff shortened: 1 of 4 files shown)


jrgriffiniii:
Past builds (please see https://circleci.com/workflow-run/62d947f1-3a67-4e9d-9066-f71bf7ba375b ) indicate that the longest steps for these jobs actually involve the usage of the Apache Solr container. I'm not certain if different jobs can share the same container, but otherwise these don't seem to consume too much time in their execution.

jrgriffiniii:
Past builds (please see https://circleci.com/workflow-run/62d947f1-3a67-4e9d-9066-f71bf7ba375b ) indicate that the longest steps for these jobs actually involve the usage of the Apache Solr container. I'm not certain if different jobs can share the same container, but otherwise these don't seem to consume too much time in their execution.

jrgriffiniii:
https://circleci.com/workflow-run/14679324-8cc0-497c-a372-63a0b1c891ef failed while https://circleci.com/workflow-run/62d947f1-3a67-4e9d-9066-f71bf7ba375b succeeded; perhaps something was cached during the Webpack build erroneously?

jrgriffiniii:
`.internal_test_app/app/assets/images` is no longer being created during the test build.

lib/generators/geoblacklight/install_generator.rb
```diff
@@ -91,7 +91,17 @@ def update_application_name
...
 
+    # Ensure that assets/images exists
+    def create_image_assets_directory
+      FileUtils.mkdir_p('app/assets/images') unless File.directory?('app/assets/images')
+    end
+
     def bundle_install
+      # Override the sass-rails problem
+      if Rails.version =~ /5\.[12]/
+        gsub_file('Gemfile', "gem 'sass-rails', '~> 5.0'", "gem 'sass-rails', '~> 6.0'")
```
mejackreed: There is a mechanism to resolve this using Gemfile.extra approach. Was that not working?

jrgriffiniii: It was not for me, I can make a second attempt.

jrgriffiniii: Trying to override the `sass-rails` release with https://github.com/geoblacklight/geoblacklight/pull/876/commits/e586b1485188053cf20a354c1259187d01ed26d5 broke for me locally. Should I instead downgrade some other dependency?

jrgriffiniii: Downgrading `sprockets` didn't break anything for me, I can get this rebased with that commit and request another review.

mejackreed: Did you `rf -rf .internal_test_app` and `rm Gemfile.lock` and regenerate?

jrgriffiniii: I apologize, I spoke too soon. The `Gemfile.extra` approach still seems to be breaking.

mejackreed: Do we need to add something like this to the `Gemfile`?

```ruby
if File.exists?('spec/test_app_templates/Gemfile.extra')
  eval File.read('spec/test_app_templates/Gemfile.extra'), nil, 'spec/test_app_templates/Gemfile.extra'
end 
```

The reason I'm advocating for this approach is for consistency with other Blacklight engine projects.

jrgriffiniii: Thank you very much, that seems to have fixed everything for me locally.

PR #880, opened 2020-01-28
Bug Fix: update copyField statements for "catch all" search field

### LINKED ISSUE #857: Solr - schema.xml - Search catch all field "text" is unpopulated.
### Descriptive summary

GeoBlacklight#master and probably older releases for quite some time. The B1G Geoportal exhibits this issue and maybe some other running GBL instances, too.

### Expected behavior

The "text" field is our catch all field for building tokens for search. As our schema.xml file is currently written, nothing is copied over into this field.

### Actual behavior

This field should hold many tokens from many fields, and act as the lowest-relevant default search field, as defined in solrconfig.xml's /select requestHandler: text^1

1. Checkout/clone GeoBlacklight#master
1. Edit the solrconfig.xml file to remove all "qf" and "pf" field boosts from the /select requestHandler, except for the text^1 entry. Save file.
1. Spin up GeoBlackight via: bundle exec rake geoblacklight:server
1. Perform a search for anything. You will return no results, because the "text" field is empty and cannot match any of your query tokens.

### Related work

I don't think this is related to anything prior, perhaps just overlooked?

ewlarson:
From Solr docs: "Copying is done at the stream source level and no copy feeds into another copy. This means that copy fields cannot be chained i.e. you cannot copy from here to there and then from there to elsewhere."

So! copying into a dynamically created wildcard field doesn't allow us to then "chain" that new dynamic field value to populate the "text" field.

Switch to using *_s and *_sm fields to ensure our "catch all" field is properly populated.

Fixes #857

### DIFF
--- a/solr/conf/schema.xml
+++ b/solr/conf/schema.xml
@@ -147,8 +147,8 @@
     <!-- Adding field type for bboxField that enables, among other things, overlap ratio calculations -->
     <fieldType name="bbox" class="solr.BBoxField"
            geo="true" distanceUnits="kilometers" numberType="pdouble" />
-    <fieldType name="pdouble" class="solr.DoublePointField" docValues="true"/>   
-   
+    <fieldType name="pdouble" class="solr.DoublePointField" docValues="true"/>
+
 
   </types>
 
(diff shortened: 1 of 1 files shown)


solr/conf/schema.xml
```diff
@@ -169,8 +169,8 @@
   <copyField source="layer_slug_s"       dest="layer_slug_ti"       maxChars="100"/>
 
   <!-- core text search -->
-  <copyField source="*_ti"               dest="text" />
```
mejackreed: Is it possible that this now would not use the text based analyzers that Solr might use on a text field?

ewlarson: I think we're fine. The values of the string fields are copied into "text" which is a text_en field, so on index all the text_en *FilterFactory rules should apply. Here's what it looks like in Solr admin:

Screen Shot 2020-01-29 at 2 01 27 PM

And indeed the 100 terms begin to show stemming, lowercase, etc.

Screen Shot 2020-01-29 at 2 04 23 PM

PR #881, opened 2020-01-29
Parallelizing the RSpec test suites and improving the caching for the Gem bundle and internal test app.

### LINKED ISSUE #789: Parallelize CircleCI builds

jrgriffiniii:
Resolves #789. I have adjusted the names of the cache keys before rebasing (they weren't as clear before), and I will restart the build in order to link to the difference in time (demonstrating the slight in gain in performance).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,8 @@ jobs:
       RAILS_ENV: test
       RACK_ENV: test
       RAILS_VERSION: <<parameters.rails_version>>
+      COVERAGE: true
+    parallelism: 4
     steps:
       - checkout
       # Update chrome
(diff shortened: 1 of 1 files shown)


jrgriffiniii:
Comparing https://circleci.com/workflow-run/4bd1eca9-dd78-4af7-a19f-ed44e2f34db8 (the build with the caches seeded) to https://circleci.com/workflow-run/e0c55f6c-93aa-4b0b-917d-3e1950398a61, I feel that there may be enough of an improvement in the speed to justify this.

PR #882, opened 2020-01-29
[WIP] Adds experimental references and downloads enabling multiple downloads

mejackreed:
A few things about this PR.

 - It enables multiple downloads in an experimental way
 - It kind of hacks around nested Solr documents unless/until Solr can support `UUIDUpdateProcessorFactory` for child documents or we want to provide UUIDs for each child. https://issues.apache.org/jira/browse/SOLR-9477?focusedCommentId=17025912&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17025912
 - We could look at changing the UUID field away from `layer_slug_s` if we want to better support this
 - I _think_ this will be extensible to the other types of services we want to provide access to
 - Has to have more than 1 value

### DIFF
--- a/app/models/concerns/geoblacklight/solr_document.rb
+++ b/app/models/concerns/geoblacklight/solr_document.rb
@@ -32,6 +32,12 @@ module Geoblacklight
       References.new(self)
     end
 
+    ##
+    # Experimental feature
+    def experimental_downloads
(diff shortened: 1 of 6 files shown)


mejackreed:
Superseded by #916

PR #889, opened 2020-03-12
Add env variable to download path

VivianChu:
The download path can be configured using the ENV variable GEOBLACKLIGHT_DOWNLOAD_PATH.  It will be easier for other app using GeoBlacklight gem to configure the download path.

### DIFF
--- a/lib/geoblacklight/download.rb
+++ b/lib/geoblacklight/download.rb
@@ -14,7 +14,7 @@ module Geoblacklight
     end
 
     def self.file_path
-      Settings.DOWNLOAD_PATH || "#{Rails.root}/tmp/cache/downloads"
+      Settings.DOWNLOAD_PATH || ENV['GEOBLACKLIGHT_DOWNLOAD_PATH'] || "#{Rails.root}/tmp/cache/downloads"
     end
 
     def file_path_and_name


lib/geoblacklight/download.rb
```diff
@@ -14,7 +14,7 @@ def file_name
     end
 
     def self.file_path
-      Settings.DOWNLOAD_PATH || "#{Rails.root}/tmp/cache/downloads"
+      Settings.DOWNLOAD_PATH || ENV['GEOBLACKLIGHT_DOWNLOAD_PATH'] || "#{Rails.root}/tmp/cache/downloads"
```
mejackreed: @VivianChu Is it possible for you to use ENV vars to set your GeoBlacklight configurations?

So your local application might have something like this:

```yml
# config/settings.yml
DOWNLOAD_PATH: <%= ENV['GEOBLACKLIGHT_DOWNLOAD_PATH'] %>
```

My only concern with changing this in the GeoBlacklight code is that all of our configurations are consolidated using the config gem and this might start to expand that.

VivianChu: @mejackreed - Thanks for the suggestion. It works fine for me.  I'll close this PR.

mejackreed: Oh thats great to hear! 

Thanks for the contribution and raising the issue. Feel free to raise any other config issues you might run into.

Welcome 👍

PR #890, opened 2020-03-16
specify generated shapefile projection

dunn:


dunn:
test failures all look like flaky http specs

dunn:
green now

config/locales/geoblacklight.en.yml
```diff
@@ -15,7 +15,7 @@ en:
       error_with_url: 'Sorry, the requested file could not be downloaded. Try downloading it directly from: %{link}'
       export_formats: 'Export Formats'
       export_link: 'Export'
-      export_shapefile_link: 'Shapefile'
+      export_shapefile_link: 'WGS 84 Shapefile'
```
mejackreed, line 18: Can we guarantee that this is a WGS 84 shapefile for all metadata?

dunn: Currently, that's how data is reprojected, when it's fetched from GeoServer.  I'm not sure if that can be changed, though.  cc @tbrittnacher

PR #901, opened 2020-06-18
Use upstream icon rendering

cbeer:
Fixes  https://github.com/sul-dlss/earthworks/issues/647


app/helpers/geoblacklight_helper.rb
```diff
@@ -84,10 +84,10 @@ def render_facet_tags(facet)
   ##
   # Returns an SVG icon or empty HTML span element
   # @return [SVG or HTML tag]
-  def geoblacklight_icon(name)
+  def geoblacklight_icon(name, **args)
```
mejackreed, line 87: I'm good with this approach. However, I also am thinking we should end up deprecating this method during this work as well. The `queue_icon_aria_label` approach was built before `blacklight_icon` has similar functionality. And I believe we now are getting double labels.

cbeer: The double titles is fixed by https://github.com/geoblacklight/geoblacklight/pull/900.

For  deprecating `geoblacklight_icon`, do you imagine just throwing  `queue_icon_aria_label` etc away, or does it  need  a backwards-compatible path?

mejackreed: I would say that since Blacklight now provides a similar, albeit a bit different functionality, we could throw  `queue_icon_aria_label` away with a deprecation warning. Perhaps we could even be shipping a major version soon as well.

Issue #873, opened 2020-01-24
Consider making dc_rights_s optional


karenmajewicz:
This is the one remaining element that is required for a record to display in GeoBlacklight. Should we consider making a record operational even if the value is not included? For example, records by default could behave as the "Public" ones do. Only restricted records would need a value. 

What problems might this cause?

kgjenkins:
According to the GBL Schema entry for [dc_rights_s](https://github.com/geoblacklight/geoblacklight/wiki/GeoBlacklight-1.0-Metadata-Elements#rights):

> This field can be set to "Public", which allows users to view and download an item, or "Restricted", which requires a user to log in to an authentication service. If there are additional licenses or rights associated with a resource, administrators will have to create a custom field in the local Solr schema.

With only two allowed values, it already is effectively a boolean value with the added overhead of making sure your values exactly matches one of the two allowed values.  If we want to make it an simple optional boolean value, it would make more sense to call it something like `restricted`.

karenmajewicz:
I like the idea of a custom field, such as restricted_b, since the value is directly tied to behavior application. That would free up dc_rights to indicate a Rights statement or URI. Using dc_rights this way would also align better with projects like DPLA. https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http%3a%2f%2fpurl.org%2fdc%2fterms%2frights

We do want to encourage more rights and usage statements in the metadata, but since we never declared a place for it, only a few projects made custom fields.

mejackreed:
One potentially useful user facing feature of `dc_rights_s` now is the facet ability for Public/Restricted. Moving this to a boolean field might change this functionality.

+1 overall to better support additional types of rights statements. Maybe that is outside or difference from dc_rights_s

andrewbattista:
I agree with the above comments. I think that the work proposed by #879 will actually take care of this, but there are important functions associated with `dc_rights_s` that will break down if it's turned into an optional field

PR #902, opened 2020-06-18
Better support for documents w/o rights data

jkeck:
Closes sul-dlss/earthworks#637


karenmajewicz:
Can someone verify what this code does for me? Is it that if the Rights field is blank or missing, the document will be treated as "Restricted?" This would have implications for how we might close out #873

jkeck:
@karenmajewicz, yes, that was the intent of this code at least, that when the data is missing a document would fallback to being considered non-public and restricted.

karenmajewicz:
@jkeck  - perfect. Thanks for the clarification!

PR #916, opened 2020-07-01
Add support for multiple downloads using current dct_references_s approach

### LINKED ISSUE #340: Support for multiple direct file downloads per layer
@dbrower mentioned at OR2015 that having multiple file downloads for a given resource would be beneficial for VecNet.

Currently, files downloads are represented in `dct_references` using the `http://schema.org/downloadUrl` uri as a key:

``` json
{
  "http://schema.org/downloadUrl":"http://stacks.stanford.edu/file/druid:bb509gh7292/data.zip"
}
```

Can we add an enhancement to this to support additional file downloads?

@dbrower can you provide some examples?

cc @stevenvandervalk

mejackreed:
This pull request aims to fix #340 in the most backwards compatible way possible. While great work has happened in investigating alternatives to this approach (#884, #882, #846) I'm wondering if this helps us unblock the way forward so that a feature like this can be used without a large data migration.

image

This aims to accomplish the goal using the simplest means possible. While it does not accomplish some of the previous goals of using a now serialized JSON for the dct_references_s, I'm hoping it is a useful feature to adopters.

### Metadata updates

dct_references_s now can take an array as a value for `http://schema.org/downloadUrl`. That ends up looking like this (unescaped):
```json
{
  "http://schema.org/downloadUrl": [
    {
      "url": "https://cugir-data.s3.amazonaws.com/00/79/50/cugir-007950.zip",
      "label": "Shapefile"
    },
    {
      "url": "https://cugir-data.s3.amazonaws.com/00/79/50/agBROO.pdf",
      "label": "PDF"
    },
    {
      "url": "https://cugir-data.s3.amazonaws.com/00/79/50/agBROO2011.kmz",
      "label": "KMZ"
    }
  ],
  "http://www.opengis.net/cat/csw/csdgm": "https://cugir-data.s3.amazonaws.com/00/79/50/fgdc.xml",
  "http://www.w3.org/1999/xhtml": "https://cugir-data.s3.amazonaws.com/00/79/50/fgdc.html",
  "http://www.opengis.net/def/serviceType/ogc/wfs": "https://cugir.library.cornell.edu/geoserver/cugir/wfs",
  "http://www.opengis.net/def/serviceType/ogc/wms": "https://cugir.library.cornell.edu/geoserver/cugir/wms"
}
```
Where `url` and `label` are required here. Regular string values for `http://schema.org/downloadUrl` will also still work. This means no current metadata practices are needing to be updated, but those wanting to take advantage of this feature can. Also the `dct_references_s` field will still need to be stringified/escaped.

See https://github.com/geoblacklight/geoblacklight/compare/multi-downloads?expand=1#diff-f386df29f180a461a31b203807e17cf6 for a modified version of the test document submitted in #884

### DIFF
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -40,6 +40,7 @@ module GeoblacklightHelper
       }
     )
   end
+  deprecation_deprecate download_link_direct: 'Use download_link_file instead'
 
   def download_link_hgl(text, document)
     link_to(
--- a/app/helpers/geoblacklight_helper.rb
+++ b/app/helpers/geoblacklight_helper.rb
@@ -13,6 +13,20 @@ module GeoblacklightHelper
     @document.references.iiif.endpoint.sub! 'info.json', 'full/full/0/default.jpg'
   end
 
+  def download_link_file(label, id, url)
+    link_to(
+      label,
+      url,
+      'contentUrl' => url,
+      class: ['btn', 'btn-default', 'download', 'download-original'],
+      data: {
+        download: 'trigger',
+        download_type: 'direct',
+        download_id: id
+      }
+    )
+  end
+
   def download_link_direct(text, document)
     link_to(
       text,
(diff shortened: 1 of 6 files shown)


Wiscmapper:
For whatever it may be worth @mejackreed, I think this approach is awesome, and we would be able to adopt with minimal fuss!

PR #918, opened 2020-07-01
Redo the search results display to be accessible

cbeer:
- Use bootstrap collapse to handle collapsing
- Make the collapse toggle a button
- Use flexbox for truncation

Addresses part of https://github.com/sul-dlss/earthworks/issues/687

Screen Shot 2020-07-01 at 09 48 11
Kapture 2020-07-01 at 9 50 14

### DIFF
--- a/app/views/catalog/_index_split_default.html.erb
+++ b/app/views/catalog/_index_split_default.html.erb
@@ -1,19 +1,27 @@
 <% # header bar for doc items in index view -%>
 <%= content_tag :div, class: 'documentHeader row', data: { layer_id: document.id, bbox: document.bounding_box_as_wsen } do %>
-  <div class='status-icons'>
-    <%= render partial: 'header_icons', locals: { document: document } %>
-  </div>
-  <h3 class="index_title col-sm-9s cosl-lg-10 text-span">
-    <span class='caret-toggle'></span>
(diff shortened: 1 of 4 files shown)


app/views/catalog/_index_split_default.html.erb
```diff
@@ -1,19 +1,27 @@
...
 <%= content_tag :div, class: 'documentHeader row', data: { layer_id: document.id, bbox: document.bounding_box_as_wsen } do %>
-  <div class='status-icons'>
-    <%= render partial: 'header_icons', locals: { document: document } %>
-  </div>
-  <h3 class="index_title col-sm-9s cosl-lg-10 text-span">
-    <span class='caret-toggle'></span>
-    <% counter = document_counter_with_offset(document_counter) %>
-    <span class="document-counter">
-      <%= t('blacklight.search.documents.counter', :counter => counter) if counter %>
+  <h3 class="index_title col pl-2">
```
cbeer, line 3: I wasn't sure if the non-cols that had the caret spilling over into the gutter was a feature or not, but I split the difference and continue to have them spill but not as much as to be inaccessible on mobile. 🤷‍♂️

mejackreed: Thanks I think it is. This looks great.

PR #920, opened 2020-07-02
Replace readmore with standard bootstrap collapse and a little sprinkling of CSS

cbeer:
Fixes https://github.com/sul-dlss/earthworks/issues/656.


mejackreed:
cc @dunn as breaking changes around this was raised previously.

dunn:
will it be treated as semver-major this time?

mejackreed:
I'm thinking something like this will get us back into semver okayness https://github.com/geoblacklight/geoblacklight/pull/924

And then we can revert that for the v3.0 release.

PR #923, opened 2020-07-06
Map accessibility

cbeer:
Fixes https://github.com/sul-dlss/earthworks/issues/669

### DIFF
--- a/vendor/assets/stylesheets/leaflet.css
+++ b/vendor/assets/stylesheets/leaflet.css
@@ -246,7 +246,6 @@
 
 .leaflet-container {
 	background: #ddd;
-	outline: 0;
 	}
 .leaflet-container a {
 	color: #0078A8;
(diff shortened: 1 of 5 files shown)


vendor/assets/stylesheets/leaflet.css
```diff
@@ -246,7 +246,6 @@
 
 .leaflet-container {
 	background: #ddd;
-	outline: 0;
```
cbeer, line 249: I don't know if this is the best idea.. but it doesn't seem to change often and there's a bunch of tickets upstream about related things.

mejackreed: Could we just override this in our styles? I could see us removing this vendor asset in favor of using the gem

cbeer: Not really; the best we can do override it with some other non-default outline which may or may not match the styles the browser actually uses.

PR #929, opened 2020-07-10
Add border to item map and space between it and WMS table

ggeisler:
This PR address two styling issues:

- The maps on the item show page don't have borders, which feel a little odd given they are so close to other page elements that do have borders (sidebar cards), and generally they would just feel more defined on the page with a subtle border (which is easy to override locally for implementers that don't want it).

- For the WMS maps, the table is jammed up directly against the bottom of the map. Perhaps this was intentional -- because selections on the map directly affect the contents of the table -- but as a relatively new GeoBlacklight user I found this confusing; for a while I thought the map and table were some kind of  single element.

Here are those two issues as seen in current `master`:

Screen Shot 2020-07-10 at 4 23 44 PM

---
And here is how this PR changes them:

Screen Shot 2020-07-10 at 4 25 12 PM

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/item.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/item.scss
@@ -7,6 +7,7 @@
 
   #viewer-container {
     [data-map="item"] {
+      border: 1px solid $gray-400;
       height: 27.5rem;
     }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/geoblacklight/modules/item.scss
```diff
@@ -7,6 +7,7 @@
 
   #viewer-container {
     [data-map="item"] {
+      border: 1px solid #ddd;
```
cbeer, line 10: Instead of hard-coding the color, maybe there's an appropriate bootstrap variable?

ggeisler: 👍 There isn't an exact match but I updated it to something close enough.

PR #934, opened 2020-07-14
Add a svg_tooltip css class to particular icons …

### LINKED ISSUE #915: Add tooltip to data type, institution, and access type icons
Currently when you hover over one of the data type, institution, and access type icons (e.g. the icons next to an item title on the item show page) if you pause long enough you see a title attribute-based "tooltip":

 
Screen Shot 2020-06-30 at 5 54 03 PM

---

This information is useful (especially for less experienced users who might not recognize all the data type icons) but the title attribute-based "tooltip" is hard to see and requires a decent pause when hovering. 

Replacing the title attribute-based "tooltip" with a more formal tooltip, such as [Bootstrap's Tooltip component](https://getbootstrap.com/docs/4.4/components/tooltips/), would make the information easier for the user to find and digest. For example:

Screen Shot 2020-06-30 at 5 45 41 PM

---

Recommended details:

- Placement: below the icon (should work for both desktop and mobile layouts where the sidebar is stacked in the main content area)

- Background color: match the Download button background color

- Use for all contexts where the current title attribute-based "tooltip" is used (search results, title area on item show page, Data relations box on sidebar of item show page):

Screen Shot 2020-06-30 at 5 46 10 PM

jkeck:
… (format, institution, access, relationship) and some JS to invoke a Bootstrap tooltip w/ the SVG's title.

Closes #915 

This commit also adds the same field collapsing behavior on the status icons in the search results so we can both get hover behavior (for the tooltip) and toggle click behavior.

### Tooltips in Results
results-icons

### DOM Manipulation on Mouse Enter/Leave
tooltip-hover-markup

### DIFF
--- a/app/views/catalog/_header_icons.html.erb
+++ b/app/views/catalog/_header_icons.html.erb
@@ -1,3 +1,3 @@
-<%= geoblacklight_icon(document[Settings.FIELDS.GEOM_TYPE]) %>
-<%= geoblacklight_icon(document[Settings.FIELDS.PROVENANCE]) %>
-<%= geoblacklight_icon(document[Settings.FIELDS.RIGHTS]) %>
+<%= geoblacklight_icon(document[Settings.FIELDS.GEOM_TYPE], classes: 'svg_tooltip') %>
+<%= geoblacklight_icon(document[Settings.FIELDS.PROVENANCE], classes: 'svg_tooltip') %>
+<%= geoblacklight_icon(document[Settings.FIELDS.RIGHTS], classes: 'svg_tooltip') %>
--- a/app/assets/javascripts/geoblacklight/modules/svg_tooltips.js
+++ b/app/assets/javascripts/geoblacklight/modules/svg_tooltips.js
@@ -1,0 +1,31 @@
+/* global Blacklight */
+
+Blacklight.onLoad(function() {
+  'use strict';
+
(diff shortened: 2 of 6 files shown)


app/views/catalog/_header_icons.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= geoblacklight_icon(document[Settings.FIELDS.GEOM_TYPE]) %>
-<%= geoblacklight_icon(document[Settings.FIELDS.PROVENANCE]) %>
-<%= geoblacklight_icon(document[Settings.FIELDS.RIGHTS]) %>
\ No newline at end of file
+<%= geoblacklight_icon(document[Settings.FIELDS.GEOM_TYPE], classes: 'svg_tooltip') %>
```
mejackreed, line 1: Is this being overwritten by https://github.com/geoblacklight/geoblacklight/pull/934/files#diff-6ca53a94b233e28fba3b527f2fe59f72R331 ?

jkeck: https://github.com/geoblacklight/geoblacklight/pull/934/files#diff-6ca53a94b233e28fba3b527f2fe59f72R331 is in the `relations_icon` helper method.  Both are calling `geoblacklight_icon` and sending in `classes`.

mejackreed: ah.. of course :)

PR #891, opened 2020-03-19
[WIP] Updates GeoBlacklight to support Rails 6

mejackreed:

PR #939, opened 2020-07-27
Add support for rails 6

jkeck:
Re-starting some of the work in #891


lib/generators/geoblacklight/install_generator.rb
```diff
@@ -34,6 +34,11 @@ def assets
       copy_file 'geoblacklight.scss', 'app/assets/stylesheets/geoblacklight.scss'
       copy_file 'geoblacklight.js', 'app/assets/javascripts/geoblacklight.js'
 
+      if Rails.version.to_i == 6
+        append_to_file 'app/assets/javascripts/application.js',
+                       "\n// Required by GeoBlacklight\n//= require geoblacklight"
```
jkeck, line 39: Looking over the previous rails 6 PR, it looks like something a bit more in-depth was done here.  Happy to do something similar if it is important.

Issue #853, opened 2019-09-11
expose GBL version


kgjenkins:
### Descriptive summary

Would it be useful to expose the GBL version in the HTML code that is produced?

### Expected behavior

When looking at another institution's site (or even our own!), I could "view source" and quickly see the GBL version being used.  An HTML comment near the top of the page would be sufficient.

andrewbattista:
I think this is a good idea. We just added an .md file in our codebase that declares the version running. Not sure if that's the best implementation, but there should be a place for this somewhere in the code.

mejackreed:
How the Blacklight demo sight does it: 

https://github.com/projectblacklight/demo.projectblacklight.org/blob/master/app/views/catalog/_home_text.html.erb#L32

ewlarson:
Need to revisit this issue. Most installs customize the home partial, so we could move this feature into the head as a meta tag.

ewlarson:
Closing via https://github.com/geoblacklight/geoblacklight/pull/1337

kgjenkins:
FYI, some public instances are now omitting this tag for security reasons.

PR #947, opened 2020-08-04
adds version number to homepage

andrewbattista:
This PR addresses #853 and exposes the current GeoBlacklight version running on the app homepage. It references the value present in the `lib/geoblacklight/version.rb` file. 

This PR also adds `.DS_Store` to the `.gitignore` file in the project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ jetty
 .internal_test_app
 # Ignore byebug history
 .byebug_history
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ jetty
 .internal_test_app
 # Ignore byebug history
 .byebug_history
+.DS_Store
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -40,4 +40,7 @@
       <%= content_tag :div, '', id: 'map', aria: { label: t('geoblacklight.map.label') }, data: { map: 'home', 'catalog-path'=> search_catalog_path , 'map-bbox' => params[:bbox], basemap: geoblacklight_basemap, leaflet_options: leaflet_options } %>
     </div>
   </div>
+  <div align='right'>
+  <p>GeoBlacklight version <%= Geoblacklight::VERSION %></p>
+  </div>
 </div>
--- a/app/views/catalog/_home_text.html.erb
+++ b/app/views/catalog/_home_text.html.erb
@@ -40,7 +40,7 @@
       <%= content_tag :div, '', id: 'map', aria: { label: t('geoblacklight.map.label') }, data: { map: 'home', 'catalog-path'=> search_catalog_path , 'map-bbox' => params[:bbox], basemap: geoblacklight_basemap, leaflet_options: leaflet_options } %>
     </div>
   </div>
-  <div align='right'>
-  <%= content_tag :h5, t('geoblacklight.home.version') %> <%= Geoblacklight::VERSION %></p>
+  <div class="text-right">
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -19,3 +19,4 @@ jetty
 .internal_test_app
 # Ignore byebug history
 .byebug_history
+.DS_Store
```
mejackreed, line 22: Rather than add platform specific ignores here, can you set this up in your local Global Gitignore?

https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

app/views/catalog/_home_text.html.erb
```diff
@@ -40,4 +40,7 @@
       <%= content_tag :div, '', id: 'map', aria: { label: t('geoblacklight.map.label') }, data: { map: 'home', 'catalog-path'=> search_catalog_path , 'map-bbox' => params[:bbox], basemap: geoblacklight_basemap, leaflet_options: leaflet_options } %>
     </div>
   </div>
+  <div align='right'>
```
mejackreed, line 43: I think we might want to take another look at how / where this is styled. Are we trying to put this within the visible column/row layout?

app/views/catalog/_home_text.html.erb
```diff
@@ -40,4 +40,7 @@
...
+  <p>GeoBlacklight version <%= Geoblacklight::VERSION %></p>
```
mejackreed, line 44: Should this also be internationalized?

andrewbattista: Thanks for these! Yeah, I am open to any / all suggestions. Was just following that example (more or less) on the test Blacklight page. And yes, the text should be internationalized. We can discuss at the stand-up or later on the the afternoon for me? 

And as for the .DS_Store thing, sure, that can be deleted.

app/views/catalog/_home_text.html.erb
mejackreed, line 43: Rather than `align='right'` can we use the positioning class from Bootstrap?

```suggestion
  <div class="text-right">
```

app/views/catalog/_home_text.html.erb
```diff
@@ -40,4 +40,7 @@
       <%= content_tag :div, '', id: 'map', aria: { label: t('geoblacklight.map.label') }, data: { map: 'home', 'catalog-path'=> search_catalog_path , 'map-bbox' => params[:bbox], basemap: geoblacklight_basemap, leaflet_options: leaflet_options } %>
     </div>
   </div>
+  <div align='right'>
+  <%= content_tag :h5, t('geoblacklight.home.version') %> <%= Geoblacklight::VERSION %></p>
```
mejackreed, line 44: Using an `h5` here gives us a heading structure problem. 

Screen Shot 2020-08-04 at 9 45 06 AM


Also lets add the version as a variable in the i18n key so that adopters could easily remove this.

```suggestion
    <%= t('geoblacklight.home.version', version: Geoblacklight::VERSION ) %>
```

config/locales/geoblacklight.en.yml
```diff
@@ -29,6 +29,7 @@ en:
       data_type: 'Data type'
       placename: 'Placename'
       subject: 'Subject'
+      version: 'GeoBlacklight version'
```
mejackreed, line 32: Now using the version variable

```suggestion
      version: 'GeoBlacklight version %{version}'
```

mejackreed:
I'm good with this, but I see some additional pending reviews. Whomever ends up 🚢 mind squashing all of the commits down?

Issue #973, opened 2020-08-18
Syntax error originating in _blacklight_overrides.scss


jazairi:
### Descriptive summary

When using the sass-rails gem (~> 5.0) in an application using GeoBlacklight versions >= 2.2, the app throws a Sass syntax error:

`Invalid CSS after "....svg') !default": expected selector or at-rule, was ".navbar-logo { ..."`

I traced the bug to `app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss`, which is missing a semicolon at the end of line 5:

`$logo-image: image_url('blacklight/logo.svg') !default`

As far as I can tell, this only occurs with sass-rails. Sassc-rails parses the CSS without any issues (as does sass-rails 6.0, which is just a sassc-rails wrapper).

Since sass-rails is deprecated, most people are unlikely to encounter this problem, but I thought I should report it in case it causes issues for anyone else.

### Steps to reproduce the behavior

1. Install version 5.1.0 of the sass-rails gem
2. Run the application
3. Attempt to load the root route

ewlarson:
Hi @jazairi - Thanks for writing up this issue! Much appreciated. If you have time to submit a PR, that would be lovely, otherwise I can get to it early next week.

jazairi:
Hi @ewlarson! Sure, I'll submit a PR today.

mejackreed:
Resolved in #976

PR #976, opened 2020-08-19
Fix syntax in _blacklight_overrides.css

jazairi:
This fixes a syntax error in `_blacklight_overrides.scss`, a missing semicolon that breaks the application when using sass-rails ~> 5.0. This might not be an issue for most people, since sassc-rails (and/or sass-rails 6.0) seems to handle missing semicolons more gracefully. We encountered it in the midst of upgrading from an earlier version of GeoBlacklight, so hopefully this will save other folks some debugging time.


mejackreed:
Thanks for the contribution! Fixes #973

Issue #988, opened 2020-12-11
Template install and regular install instructions are broken for new users


mejackreed:
### Descriptive summary

The template based and normal full intsructions for users creating a GeoBlacklight application appear to be causing issues.

The template errors are due to a bad setup in the template.

Other issues stem from application.css strange caching issues. Originally reported by @mapninja

### Expected behavior

Our tutorials should work.

### Actual behavior

They present users with unresolvable errors.

### Steps to reproduce the behavior

Follow the instructions here: https://geoblacklight.org/tutorial/2015/02/09/create-your-application.html using <6.1 rails.

mapninja:
I've commented out EVERYTHING from `app/assets/config/stylesheets` and I get the unstyled GBL page, successfully. Uncommenting anything in /stylesheets, in any combination I've tried, causes chaos. Here's errors from the default/uncommented results:
![image](https://user-images.githubusercontent.com/1517244/101947102-c32b4400-3ba4-11eb-8d6c-ebc6121f592d.png)

andrewbattista:
Next steps: support Sprockets instead of WebPacker as an approach to enable users to spin up new instances of GeoBlacklight

Remove existing WebPacker generator in core codebase, for clarity and functionality

Establish a pipeline for determining when and why to implement WebPacker, and document it

eliotjordan:
Can we close this?

dl-maura:
I was able to use the template with Rails 6.1 and ruby 2.7, but it does NOT work on Ruby 3. Not sure if we want that to be a new ticket or use that as part of this?

eliotjordan:
I think the resolution of this ticket is to specify the rails and ruby versions users should have installed. I was able to run it with Rails 6.1 and Ruby 2.6.5.

ewlarson:
Okay! After a bunch of trial and error, here is a modified `template.rb` file, for Ruby v2.7.5. It *does not* work for Ruby v3.

With Ruby 2.7, try running:
```bash
DISABLE_SPRING=1 rails new gbl-from-template -m https://gist.githubusercontent.com/ewlarson/439ca314f202a34d209b649457b1aa19/raw/10388a4a1d66dfc1d0debbbcdedcfb6ff3f3a519/gbl_template.rb
```

https://gist.github.com/ewlarson/439ca314f202a34d209b649457b1aa19
```ruby
# frozen_string_literal: true

# Tested with Ruby version 2.7.5
gem 'blacklight', '7.22.0'
[...]

PR #998, opened 2021-02-10
Update template to make sure we use sprockets less than 4.0 until we …

mejackreed:
…can support this automaticall

When following various installation guides, our "webpacker" install doesn't work out of the box. Make sure to at least make the template usable until we can do a better job internally to generate this functionally with Sprockets > 4.0 out of the box.


mejackreed:
Partly addresses #988

eliotjordan (approved):

PR #1008, opened 2021-02-17
Bug fix: Remove JS map from bookmarks view

### LINKED ISSUE #246: Moving map in Bookmarks view gives new search results
When I view my bookmark list, if I move the map, it replaces my bookmarks with a new set of results. Expected behavior is to show only the bookmarks within the region I zoomed into.

ewlarson:
Make the GBL leaflet map selector more specific to prevent the map from rendering on the bookmarks view.

Fixes #246

### DIFF
--- a/app/assets/javascripts/geoblacklight/modules/results.js
+++ b/app/assets/javascripts/geoblacklight/modules/results.js
@@ -8,7 +8,7 @@ Blacklight.onLoad(function() {
     });
   }
 
-  $('[data-map="index"]').each(function() {
+  $('body.blacklight-catalog [data-map="index"]').each(function() {
     var data = $(this).data(),
     opts = { baseUrl: data.catalogPath },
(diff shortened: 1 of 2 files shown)


mejackreed (changes requested):
I wonder if we could/should tackle this issue a different way?

I say this as I would consider the map view of bookmarks a feature and removing this functionality would be a breaking change.

Would it be possible to instead know the context and pass an attribute through that could disable the map from displaying this?

ewlarson:
Hmmm... I thought we were leaning toward removing the bookmarks map altogether, as its current behavior is extremely unexpected. Did this work correctly in the past?

To have the map pan/zoom search work, bookmarks would need to be searchable. Or, are we talking about removing the geosearch functionality of the map?

mejackreed:
I'm talking specifically about removing the ability for the map to searchable in the context of the `BookmarksController`

ewlarson:
Closing. After discussion, I'll try another path.

PR #1016, opened 2021-02-19
Add support for WMTS layers in GeoBlacklight fixes #898

### LINKED ISSUE #898: Add WMTS capability
Suggest we consider adding OGC Web Map Tile Service to the list of allowable external references. (and extending GBL as needed, of course)  

Here is our use case: we catalog and provide access to thousands of downloadable image mosaics and individual DOQQ image tiles. It would be nice to show a WMTS service is the background, instead of just the footprint overlaid on the standard basemap. 

Rationale: an image background (served via WTMS) provides context to the user so they have a better sense for what they are downloading.

The status quo, no WMTS, standard basemap visible:

image

I hacked wms.js to mock up an example of the desired end result:

image

Zoomed out, notice how the WMTS overlays on standard basemap, much like we can currently do with WMS.  I also wanted the footprint to be visible as well, which is NOT the default functionality for WMS:

image

mejackreed:
If this is accepted, happy to update the wiki documentation with the new dct_reference

https://github.com/geoblacklight/geoblacklight/wiki/External-references

Kapture 2021-02-19 at 08 20 19

A similar pattern could be done for TMS if that is a needed data source

++ @Wiscmapper for providing the fixture which made this nice to implement


eliotjordan (changes requested):
@mejackreed This looks good except I think this implementation and the fixture document are really TMS rather than WTMS. I know the difference is somewhat arbitrary given the similarity between the two, but I think that the WTMS spec requires that tiles can be requested via REST and KVP encodings. This ESRI documentation (of all sources) has a clear example of both styles: https://developers.arcgis.com/rest/services-reference/wmts-tile-map-service-.htm I wonder if this should just be TMS and we can implement WMTS if specifically asked for later on. An example plugin if we want to do that: https://github.com/alexandre-melard/leaflet.TileLayer.WMTS

eliotjordan:
I have some reservations about my previous comment. Does `WTMS == z/y/x` and `TMS == z/x/y`?

eliotjordan:
TMS Spec for reference and for use as a reference key if needed: https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification

mejackreed:
I'm generally confused about the difference between TMS and WMTS now, and wonder if we actually have any WMTS data then that we can reference.

mejackreed:
So it seems like we will definitely need a bigger effort than this to get WMTS supported. I'm going to unassign myself and convert this to a draft PR for future reference.

The work here needing to be:
 - Adopt or write a Leaflet WMTS layer plugin ( I had some issues w/ version compatibility with https://github.com/alexandre-melard/leaflet.TileLayer.WMTS)
 - Pass through the correct parameters to the plugin

mejackreed:
Closing in favor of #1030

PR #1017, opened 2021-02-19
Remove Teaspoon dependency in favor of Jest

### LINKED ISSUE #680: Reevaluate testing and integration for JavaScript

mejackreed:
While teaspoon was nice, it seems that there are some maintenance issues upstream and limits Ruby 3.0 ability to test in CI.

Fixes #680


ewlarson (approved):
Works lovely after a `yarn install`.

PR #1030, opened 2021-02-24
Adds support for a TMS viewer

mejackreed:
In place of #1016

Screen Shot 2021-02-24 at 7 41 11 AM

TMS layers would be specified like this:

```
"https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification":"https://earthquake.usgs.gov/basemap/tiles/faults/{z}/{x}/{y}.png"
```

This allows you to modify the tile pattern if need be.

### DIFF
--- a/app/assets/javascripts/geoblacklight/viewers/tms.js
+++ b/app/assets/javascripts/geoblacklight/viewers/tms.js
@@ -1,0 +1,10 @@
+//= require geoblacklight/viewers/wms
+
+GeoBlacklight.Viewer.Tms = GeoBlacklight.Viewer.Wms.extend({
+
+  addPreviewLayer: function() {
+    var _this = this;
+    var wmtsLayer = L.tileLayer(this.data.url);
+    this.overlay.addLayer(wmtsLayer);
+  }
+});
--- a/config/locales/geoblacklight.en.yml
+++ b/config/locales/geoblacklight.en.yml
@@ -64,6 +64,7 @@ en:
       data_dictionary: 'Documentation'
       services: 'Web services'
       services_close: 'Close'
+      tms: 'Tile Map Service'
     relations:
       ancestor: 'Source Datasets'
       descendant: 'Derived Datasets'
@@ -104,6 +105,9 @@ en:
         tiled_map_layer:
           title: ArcGIS Tiled Map Layer
           content: An ArcGIS Tiled Map Layer Service displays set of web-accessible tiles that reside on a server.
+        tms:
+          title: Tile Map Service
+          content: A Tile Map Service displays georeferenced map tiles as an image on a map.
         wms:
           title: Web Map Service (WMS)
           content: A Web Map Service displays a geospatial dataset as map images.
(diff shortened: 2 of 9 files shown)


ewlarson:
@mejackreed / I've got time to review PRs if you can fix this one broken test.

```ruby 
Failures:

  1) Index view should have documents and map on page
     Failure/Error: expect(page).to have_css('.document', count: 5)
       expected to find visible css ".document" 5 times, found 6 matches: "1.\n10 Meter Contours: Russian River Basin, California", "2.\n1-Meter Shaded Relief Multibeam Bathymetry Image (Color): Elkhorn Slough, California, 2005", "3.\n2005 Rural Poverty GIS Database: Uganda", "4.\nDabao Kinbōzu, Maps Index", "5.\nJōshū Kusatsu Onsenzu", "6.\nQuaternary Fault and Fold Database of the United States"
     # ./spec/features/split_view.html.erb_spec.rb:13:in `block (2 levels) in <top (required)>'

Finished in 4 minutes 42.4 seconds (files took 4.59 seconds to load)
337 examples, 1 failure, 5 pending
```

ewlarson (approved):
Tusen takk.

mejackreed:
What would we use as the URI in dct_references?

Wiscmapper:
@karenmajewicz I believe the answer to your question is, I think so.  This topic evolved pretty quickly back in mid-February before I had a chance to comment.  And I hadn't given this subject much thought since last May.  I think the addition of a tile layer (aka xyz) capability is awesome, and helps. So go forth!

I don't suggest any change of direction to what's been done for this release, simply adding some comments for future reference.

I do think the notion of WMTS, which is a different beast, could be revisited someday. For anybody in the future who might read this thread and want to revisit, read the following in this order:

https://en.wikipedia.org/wiki/Tile_Map_Service
"Tile Map Service or TMS, is a specification for **_tiled web maps_**, developed by the Open Source Geospatial **Foundation**"

Then read:
https://en.wikipedia.org/wiki/Web_Map_Tile_Service
"A Web Map Tile Service (WMTS) is a standard protocol for serving pre-rendered or run-time computed georeferenced map tiles over the Internet. The specification was developed and first published by the Open Geospatial **Consortium** in 2010 ... **An earlier specification for this is the Tile Map Service (TMS). It is simpler than WMTS.** It was developed by members of the OSGeo and **[TMS] is not backed by an official standards body.**"

Lastly:
https://en.wikipedia.org/wiki/Tiled_web_map
"The de facto OpenStreetMap standard, known as Slippy Map Tilenames[2] or XYZ ..."

**So if you are keeping score:**
If you believe Wikipedia (?) TMS is sort of an xyz tiled web map, which is why I think it's horribly confusing.  I think it's best just to forget about TMS altogether!
TMS and WMTS were developed by different organizations
WMTS is newer, it seems
WMTS is a well-defined standard supported by an international organization
Simple XYZ tiles are useful, and are a git 'er done sort of approach to consuming pre-generated raster tiles with minimal fuss and overhead

PR #1035, opened 2021-02-25
Use bash to run yarn test command

eliotjordan:
Jest tests were not running on github actions CI:

```
yarn run v1.22.5
$ jest -c jest.config.js
/bin/sh: 1: jest: not found
error Command failed with exit code 127.
```

The solution is to force running the command with bash instead of sh (https://github.com/yarnpkg/yarn/issues/6686)


ewlarson (approved):
👍

PR #1036, opened 2021-02-25
Make the overlap ratio field configurable

eliotjordan:


ewlarson (changes requested):
Nice catch! Can you please add the new OVERLAP_FIELD field setting to this file too:
https://github.com/geoblacklight/geoblacklight/blob/main/config/initializers/new_gbl_settings_defaults_3_3.yml

This yml file helps existing GBL installs load all the new Settings.FIELDS values if they aren't present in their local settings.yml file.

eliotjordan:
@ewlarson Done.

ewlarson (approved):

PR #1049, opened 2021-05-24
Provide a more graceful fallback for non-parsable geometry

### LINKED ISSUE #1042: Provide a more graceful fallback for non-parsable geometry
### Descriptive summary

Some records may not be parsable by RGeo, and now raise an error. We should do something better so there aren't errors in search results. https://github.com/geoblacklight/geoblacklight/blob/main/lib/geoblacklight/geometry.rb#L20

Downstream issues: https://github.com/sul-dlss/earthworks/issues/778

eliotjordan:
Closes #1042 

My comment from the ticket:

> The underlying issue here, is that the envelopes / bounding boxes should be valid polygons. The most important reason is to enable a valid solr spatial search with the record. Perhaps we should add some validation to the envelope in GeoCombine. That being said, invalid polygon geometry should not prevent record display in GeoBlacklight. After doing some reading and experimentation, I've come to understand that this is a bit of a tricky issue. As far as I can tell, the problem is the underlying version of libgeos, which can be different between machines. For example, I can't replicate this directly on my development laptop. Everything works as it should. I'll work up a patch that tests the error indirectly and let's see if it fixes the problem.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -113,7 +113,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [2.5]
+        ruby: [2.6]
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


ewlarson (approved):
This looks good. 👍

PR #1050, opened 2021-05-26
Return default GeoJSON extent on error

### LINKED ISSUE #1042: Provide a more graceful fallback for non-parsable geometry

eliotjordan:
Closes #1042 

Provides a default extent value when there is an error parsing a geometry or transforming into GeoJSON.

### DIFF
--- a/lib/geoblacklight/geometry.rb
+++ b/lib/geoblacklight/geometry.rb
@@ -19,7 +19,7 @@ module Geoblacklight
       RGeo::GeoJSON.encode(obj).to_json
     rescue StandardError
       Geoblacklight.logger.warn "Geometry is not valid: #{geom}"
-      ''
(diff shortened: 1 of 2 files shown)


ewlarson (approved):
👍

karenmajewicz:
Commenting to clarify that the default error used is a world extent [-180.0, 90.0], [-180.0, -90.0], [180.0, -90.0], [180.0, 90.0], [-180.0, 90.0]

Issue #1052, opened 2021-08-11
Determine default metadata fields to display on item view pages


karenmajewicz:
The default interface includes a short set of metadata fields that display on the item view pages. This issue proposes that we review this list and come up with a new set related to the Aardvark schema.

karenmajewicz:
I think that this doc controls what appears on the item view page: https://github.com/geoblacklight/geoblacklight/blob/7d3c31c7af3362879b97e2c1351a2496c728c59c/lib/generators/geoblacklight/templates/catalog_controller.rb

Here are what the names of the fields are and the shortnames for the URIs:

- 'Author(s)', itemprop: 'author'
- 'Description', itemprop: 'description', helper_method: :render_value_as_truncate_abstract
- 'Publisher', itemprop: 'publisher'
- 'Collection', itemprop: 'isPartOf'
- 'Place(s)', itemprop: 'spatial', link_to_facet: true
- 'Subject(s)', itemprop: 'keywords', link_to_facet: true
- 'Year', itemprop: 'temporal'
- 'Held by', link_to_facet: true

karenmajewicz:
Other fields to potentially add:

- [ ] Resource Class
- [ ] Resource Type
- [ ] ISO Topic
- [ ] Date Issued
- [ ] Rights
- [ ] Rights Holder
- [ ] License
- [ ] Format
- [ ] File Size
- [ ] Georeferenced
- [ ] Language

Fields with a potential name change:

- 'Held by' -> Provider

karenmajewicz:
I am not sure why I closed this, but I think it would be good to add more fields to the default layout.

Issue #1041, opened 2021-04-05
Determine the icon behavior for GeoBlacklight with Aardvark for Resource Class and/or Type


karenmajewicz:
Currently, an icon is displayed based upon the value in Geometry Type. The new metadata schema deprecates Geometry Type and proposes 2 new fields: Resource Class and Resource Type. See issue #1000 

**Resource Class** (Maps, Datasets, Web services, Websites, Collections) is a required field for the new schema. Some of the existing geometry type icons (Image, Collection) could be migrated here.

**Resource Type** is optional and has a much larger set of controlled values that can be thought of as subcategories for Resource Class. For datasets, this could be "Point data", "Raster data", etc. For scanned maps, this might be something like "Topographical maps". Note that Point, Line, Polygon, and Raster from Geometry Type can be migrated here, but should now have the word "data" after it. This is to better match the plural format of the other terms and to clarify to users that the terms refer to GIS data.

I am curious to hear more input on the desire for retaining the point, line, polygon icons. Since Aardvark has been designed for a larger variety of items, it was harder to work in a single field for a strict vocabulary of geometry types. One option could be to drop these icons and replace them with a single "Datasets" icon based upon Resource Class.

Alternatively, if GBL administrators want to maintain the icons for Point, Line, Polygon, etc., what would be the best strategy here?  Since both Resource Class and Type will be multi-valued, could GeoBlacklight be configured to read the first value to determine the icon?  (I believe that is how the OpenGeoPortal schema worked.) Otherwise, is a separate element still needed just for display purposes?

karenmajewicz:
To be discussed at next sprint

karenmajewicz:
Decision is to make these optional for now. Code needs to be written that will take into account multivalued fields.

PR #1074, opened 2021-08-27
Adds Aardvark fields

karenmajewicz:
This is for reference only - not for merging! This is supporting documentation for issue #1052  I think this is the easiest way to see all of the remaining Aardvark fields that need to be incorporated into the code.

All literal fields (those without URIs) have been added to settings.yml. Fields that might be added to an item view pages are in catalog_controller.rb.

Issues to sort out:

- This layout uses the change from 'provenance' to 'provider' (see #1062 for more places where this would need to be changed)
- This updates the terms for the mandatory field `dct_accessRights_s`. This field should be described as ACCESS_RIGHTS throughout the application for clarity. The new field `dct_rights_sm` should have the term RIGHTS
- This removes the confusing term GEOM_TYPE and assigns `gbl_resourceType_sm` to the TYPE term. Changing this will have implications for the icon. (see #1041)


ewlarson:
@karenmajewicz - Okay! I think is is ready to go. @andrewbattista  - Tagging you here if you have time to review this PR.

andrewbattista (approved):
This looks good to me. I appreciate the congruence between the schema and the code that now exists. Thanks to @karenmajewicz and @ewlarson

PR #1078, opened 2021-08-31
Deprecation Fix - presenter

### LINKED ISSUE #1067: DEPRECATION WARNING: #presenter
### Descriptive summary

CatalogController#index as HTML (search results) / Appears 10 times on page load

DEPRECATION WARNING: #presenter is deprecated; use #document_presenter instead. (called from geoblacklight_present at /Users/ewlarson/Rails/geoblacklight/lib/geoblacklight/geoblacklight_helper_behavior.rb:12)

ewlarson:
Fixes #1067 

Moves GeoBlacklightHelperBehavior from lib to app/helpers, which is better aligned with Blacklight's HelperBehavior files, and allows us to include Blacklight::BlacklightHelperBehavior to access the document_presenter method.

Also adds skips for specs that reference Purdue ArcGIS web services, which are returning CORS errors at the moment.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -65,7 +65,7 @@ Lint/BooleanSymbol:
 # AllowedMethods: enums
 Lint/ConstantDefinitionInBlock:
   Exclude:
-    - 'spec/lib/geoblacklight/geoblacklight_helper_behavior_spec.rb'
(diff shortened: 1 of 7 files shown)


tampakis (approved):

dunn:
this change depends on https://github.com/projectblacklight/blacklight/commit/43f2d324787eaa201a59af65b9dff04669753ac2 which means that geoblacklight now requires blacklight 7.12 or higher

PR #1087, opened 2021-09-03
Homepage map: use map-geom not map-bbox

### LINKED ISSUE #1040: Homepage Map - Incorrect data-map-bbox attribute in v3.3.0
### Descriptive summary

In GBL v3.3.0, we added support for complex geometries (#1032) and changed the map data attribute from map-bbox to map-geom, to indicate support for much more than simple bboxes. However, we forgot the homepage map also used the data-map-bbox attribute to potentially scope the initial map view from world to something pertinent to the local collection.

### Expected behavior

Homepage map would use the data-map-geom attribute. 

Example v3.3.0 use to scope the homepage map via the new data-map-geom attribute, via a GeoJSON object:
https://github.com/umass-gis/geoblacklight/blob/main/app/views/catalog/_home_text.html.erb#L40

### Actual behavior

Out of the box, GBL's `catalog/_home_text.html.erb` file still references the now unused data-map-bbox attribute.

### Related work

* Support for complex geometries (#1032)

ewlarson:
The new complex geometries feature changed the data-map attribute from data-map-bbox to data-map-geom, to indicate support for GeoJSON objects instead of simple bboxes.

No spec changes to accompany this change, as the homepage default behavior is to not have a geom (or GeoJSON) boundary, and without a param value here ERB strips the attr from the output.

Example homepage map boundary customization here:
https://github.com/umass-gis/geoblacklight/commit/0cd2093a851bdc5afa4ab224e5e697b93c9f5526

Fixes #1040


tampakis:
Given the example that's provided, I think this seems like something you should be able to configure in the settings.yml file so this file won't necessarily need to be overriden in the local app

ewlarson:
Good idea! I'll add that option here...

tampakis:
This is great and now there is a test and some in-line documentation!

tampakis (approved):

ewlarson:
One extra commit to fix the comment on rails_config.rb

tampakis:
Sounds good, merge away after CI passes!

PR #1091, opened 2021-09-09
Added French localization and added some English fields

pdante-ubc:
Added French localization we had translated for the Geodisy project. Had to add some additional yml files to correspond to some of the blacklight and other files that also needed translating and added a few new English fields as well. Please let me know if you have any questions.


eliotjordan:
@pdante-ubc We'd love to have this in the codebase. It needs a rebase to solve the stalled CI problem. More importantly, there have been quite a few changes since last September and I don't know if they have any impact on these settings. Let me know if I can help with this.

ewlarson:
Hi @pdante-ubc and @eliotjordan:

I'm also a fan of this PR in theory, but I think its current structure would be difficult for us to maintain in GeoBlacklight:

* Some files would be best contributed upstream to Blacklight proper: `config/locales/blacklight.en.yml` and  `config/locales/blacklight.fr.yml`
* Some files appear unnecessary: `config/locales/en.yml` and `config/locales/fr.yml` -- both of these files have translations that are "blacklight" namespaced, so these keys could be in the blacklight i18n files.
* Also, I think these should be `fr-CA.yml` translations and not `fr.yml` -- Gimlet is translated into both Français (Canadien, fr-CA) and Français (fr) and our Canadian clients disagree with our French clients on quite a few translations.

Maybe we could all meet online to chat about a good path forward here? Perhaps this would be a good topic for an upcoming GeoBlacklight community call? March 9th at 1:05 – 1:55pm (CST) is the next one.

Thoughts?

eliotjordan:
I'm closing this as there hasn't been activity since 2022. Localizations are great. If there is energy to do this, I'd be happy to help.

Issue #1104, opened 2021-10-28
WMS params WIDTH and HEIGHT should be integers


kgjenkins:
### Descriptive summary

We've run into an issue when clicking on a map feature in Firefox occasionally fails to fetch the feature info from GeoServer via a WMS GetFeatureInfo request, which reports an error like:
```
java.lang.NumberFormatException: For input string: "604.23"
For input string: "604.23"
```

According to the OGC WMS standard, the WIDTH and HEIGHT parameters should be integers, but Firefox is sometimes using floating-point numbers.  This tends to happen if the user is zoomed into the page to 110% (or any value other than 100%).

GBL is using jQuery to get these values as follows ([source code](https://github.com/geoblacklight/geoblacklight/blob/main/app/assets/javascripts/geoblacklight/viewers/wms.js#L43-L44)):
```
WIDTH: $('#map').width(),
HEIGHT: $('#map').height(),
```

But the jQuery docs say that "The number returned by dimensions-related APIs, including .width(), may be fractional in some cases. Code should not assume it is an integer."

### Proposed solution

I'll make a pull request to explicitly convert these values to integers.

karenmajewicz:
pull request submitted - merge when CI is fixed

eliotjordan:
Closed by https://github.com/geoblacklight/geoblacklight/pull/1105

PR #1105, opened 2021-10-28
WMS params WIDTH and HEIGHT should be integers

kgjenkins:
Fixes issue #1104


ewlarson:
@eliotjordan / I looked into the Rails 5.2 failure here a bit, but wasn't able to see anything obvious that causes the error -- there's a webpacker gem version difference between the 6.0 and 5.2 builds, but that doesn't appear to be the cause.

With the CI suite failing, we're stuck here unless we want to press forward. Any concerns if we merge this without a 5.2 resolution?

Also! I think we should drop Rails 5.2 support (CI testing) from the v4 branch altogether. I think we can safely suggest Rails 6.1 for all GBLv4 adopters.

PR #1111, opened 2022-01-14
Add XYZ tiles reference

eliotjordan:
- Add reference and viewer for XYZ tiles
- Removes TMS reference and viewer
  - I really struggled to find a working TMS example out the wild. It doesn't appear to be very common.
  - Also, TMS can often be converted to XYZ by negating the z value in the url
- Issue with latest solr (Could be split in to new PR)
  - Fix an issue in the solr config with Solr 8.11.
- Updates to CI (Could be split in to new PR)
  - Pin bundler at 2.1.1 until https://github.com/cbeer/engine_cart/issues/106 is resolved.
  - Test latest rails versions: 6.1 and 7.0.

It appears to me that we've had a lot of problems with Ruby v3 lately. I couldn't get tests to run, so I've removed it from the CI matrix and will open a ticket to investigate.


ewlarson:
I'm deep into another project at the moment, but just a few thoughts here:

1. Correct me if I'm wrong, but the tms method should be marked deprecated not removed entirely (could remove entirely in v4?)
2. Seems Ruby 3.1 is the one failing, not Ruby 3.0 (I think... [more here](https://bibwild.wordpress.com/2021/12/28/github-action-setup-ruby-needs-to-quote-3-0-or-will-end-up-with-ruby-3-1/) ... we need to quote '3.0' for 3.0)
3. I'd like to see the Solr and CI updates in separate PRs

Not trying to be a blocker here. Just think this PR looks like a few good PRs in one.

eliotjordan:
@ewlarson 

1. TMS is easy enough to support. I'll just leave it be and we can work on finding a valid real-world service for our fixtures: https://github.com/geoblacklight/geoblacklight/issues/1119
2. Good catch. I'll add the quote.
3. Done

eliotjordan:
Closing this PR because I a have it into three smaller PRS:
https://github.com/geoblacklight/geoblacklight/pull/1114
https://github.com/geoblacklight/geoblacklight/pull/1118
https://github.com/geoblacklight/geoblacklight/pull/1117

Issue #1110, opened 2022-01-13
Add XYZ reference and viewer


eliotjordan:
- Also should fix the tms viewer which is incorrect.
- Backport to version 3.0

PR #1118, opened 2022-01-24
Add XYZ tiles reference

eliotjordan:
Works towards #1110 

Created follow-on issue: https://github.com/geoblacklight/geoblacklight/issues/1119


thatbudakguy (approved):
i think i mostly understand what's going on here 🙂

PR #1125, opened 2022-01-31
Developer improvements

### LINKED ISSUE #1123: Add .tool-versions file for asdf users
Some developers use asdf instead of rvm, and this requires using a .tool-versions file.

### LINKED ISSUE #1124: Can't run solr wrapper without admin access on computer
### Descriptive summary

Without admin acess, solr wrapper won't start.

### Expected behavior

Solr wrapper starts

### Actual behavior

Solr never starts

XanderGardner:
- Added .tool-versions file for users who use asdf instead of rvm. Closes #1123 
- Set solr instance directory to tmp/solr so that users without admin access of computers can run solr. Closes #1124


.tool-versions
```diff
@@ -0,0 +1,3 @@
+ruby 2.7.5
+nodejs 17.4.0
+java openjdk-11.0.2
```
thatbudakguy, line 3: in my experience, most projects like geoblacklight tend not to commit these types of files. the reasons (i think) are twofold:
- doing so would result in a proliferation of files for all the various "version managers", e.g. rbenv people have `.ruby_version`, asdf people have `.tool-versions`, etc. then when you want to update a version, you need to make sure to update all of them. and, if someone wants to use a (supported) ruby version other than 2.7.5, they need to override this file.
- i think the project actually supports a range of versions and intentionally doesn't specify any specific ones (at least according to the [development wiki page](https://github.com/geoblacklight/geoblacklight/wiki/Development)). so while these versions are valid, there are many other combinations that are equally valid, and we don't want to tell people to do it one particular way.

hope i've correctly understood what this file is for; if not this might not make sense. here's [an example where GitLab chose to ignore these files](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/35497), although there seem to also be others where people chose to commit them.

XanderGardner: Makes sense and thank you! I added .tool-versions to the gitignore file instead. I moved my work to the pr #1126

XanderGardner:
PR moved to #1126

PR #1130, opened 2022-02-04
Add a viewer for TileJSON and WMTS endpoints

### LINKED ISSUE #1129: Add reference and viewer for TileJSON endpoint
TileJSON Spec: https://github.com/mapbox/tilejson-spec

eliotjordan:
- Adds TileJSON viewer
- Adds WMTS viewer
- Use WebMock gem to mock WMTS and TileJSON manifest requests.

Closes #1129 

Screen Shot 2022-02-03 at 10 23 43 PM

Screen Shot 2022-02-03 at 10 24 45 PM

Screen Shot 2022-02-04 at 7 14 34 PM

### DIFF
--- a/app/assets/javascripts/geoblacklight/viewers/wmts.js
+++ b/app/assets/javascripts/geoblacklight/viewers/wmts.js
@@ -1,0 +1,85 @@
+//= require geoblacklight/viewers/wms
+
+GeoBlacklight.Viewer.Wmts = GeoBlacklight.Viewer.Wms.extend({
+
+  addPreviewLayer: function() {
+    var _this = this;
+    fetch(this.data.url).then(function(response) {
+      if (!response.ok) {
+          throw new Error("Unable to fetch WMTSCapabilities document");
+      }
+      return response.text();
+    }).then(function(wmtsCapabilities) {
+      var tileInfo = _this.parseCapabilities(wmtsCapabilities),
+          options = tileInfo.bounds ? { bounds: tileInfo.bounds } : {},
+          tileLayer = L.tileLayer(tileInfo.url, options);
+      _this.overlay.addLayer(tileLayer);
+    }).catch(function(error) {
+      console.debug(error);
+    });
+  },
+
+  parseCapabilities: function(wmtsCapabilities) {
+    var parser = new DOMParser(),
+        doc = parser.parseFromString(wmtsCapabilities, "application/xml"),
+        layers = doc.getElementsByTagName("Layer"),
+        selectedLayer = this.findLayer(layers),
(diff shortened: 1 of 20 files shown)


app/assets/javascripts/geoblacklight/viewers/wmts.js
```diff
@@ -0,0 +1,85 @@
...
+    var parser = new DOMParser(),
+        doc = parser.parseFromString(wmtsCapabilities, "application/xml"),
+        layers = doc.getElementsByTagName("Layer"),
+        selectedLayer = this.findLayer(layers),
+        resourceElement = selectedLayer.getElementsByTagName("ResourceURL")[0];
+
+    // Generate URL with Leaflet.TileLayer URL template formatting
+    var tileURL = resourceElement.getAttribute("template")
+      .replace("{Style}", this.getStyle(selectedLayer))
+      .replace("{TileMatrixSet}", this.getTileMatrixSet(selectedLayer))
```
thatbudakguy, line 32: are there any negative consequences if either of these are `undefined` (because there was no `<Style>` or `<TileMatrixSet>`)? can that happen?

thatbudakguy (approved):
looks good to me!

tangentially, it seems like we're adhering to pretty old-school javascript conventions (`for` loops with `i`, using `var` everywhere, etc). i thought that might be for IE compatibility, but then we're also using `fetch`, which isn't in IE. this is probably a broader blacklight thing, but if we don't care as much about IE, it'd be nice to use `let`/`const` and the more idiomatic stuff for iterating like `Array.prototype.forEach`. just my $0.02

app/assets/javascripts/geoblacklight/viewers/wmts.js
eliotjordan: It's really up to the creator of the WMTSCapabilites document. `Style` and `TileMatrixSet` are required according to the spec, but sometimes the URL template doesn't make use of it. The [Princeton WMTSCapabilities](https://github.com/geoblacklight/geoblacklight/pull/1130/files#diff-18211e325f3cabced0e3dd77e290670f4d639591d364cd246e88482ad78d5ebdR51) fixture is an example. If the template requires them and they are aren't defined, then adding the layer will fail. Which is expected.

eliotjordan:
@thatbudakguy Your comments about javascript style are valid. I had the same thoughts, but wanted to match the existing patterns and style. I created a ticket to look at the this as a group. https://github.com/geoblacklight/geoblacklight/issues/1132

app/assets/javascripts/geoblacklight/viewers/wmts.js
thatbudakguy: Makes sense!

PR #1140, opened 2022-02-14
Fix coverage badge and simplify coverage checking

### LINKED ISSUE #1139: Fix coverage badge and add simplecov coverage protection
### Descriptive summary

The badge link is broken. We should fix that and add a simpler simplecov coverage protection. 

### Related work

Example here:
https://github.com/samvera-labs/geoserver-publish/blob/main/README.md?plain=1#L4
https://github.com/samvera-labs/geoserver-publish/blob/main/spec/spec_helper.rb#L13

eliotjordan:
Closes #1139


ewlarson (commented):
@eliotjordan / The simplecov changes look good here. If that new coverage badge is always going to say 100% green (looks like it doesn't "know" our coverage), maybe we don't need a coverage badge for the project?

hackartisan:
I like the badge even if it doesn't change. It's a form of documentation.

PR #1166, opened 2022-02-17
Copy leaf and branch icons to have new titles

### LINKED ISSUE #935: Update <title> labels for Data relations icons
#910 enables using the format type icon for items shown in the Data relations box, rather than the previous leaf and branch icons. However, an implementer has the option of using the leaf and branch icons instead of the format type icon. But the current SVG `<title>` labels for the leaf and branch icons are not very informative to the user:

Screen Shot 2020-07-14 at 10 53 37 AM

I suggest updating the title labels for these two SVGs:

- `leaf` becomes `Child item`
- `pagelines-brands` becomes `Parent item`

(Alternatively, we could use `Child record` and `Parent record` to be more consistent with the within-card labels that use the term `record` but I personally feel that "record" is a term more appropriate for librarians than non-librarian end-users.)

Those labels work better for collection-type relationships than they do for derived-type of relationships, but they are at least an improvement on the current labels, which are just the source names of the icons and not informative to the user.

hackartisan:
Should we note this in the upgrade guide? If folks want this change they'll have to update their local settings.yml.

closes #935


ewlarson (approved):
Looks good. I'll start a list of the settings.yml changes for GBL v4 on the upgrade guide.

PR #1167, opened 2022-02-17
Update configuration, copy fields, and specs / comments

### LINKED ISSUE #1155: Update solrconfig.xml with new Aardvark field names
### Descriptive summary

These fields need to be updated with Aardvark field names.
https://github.com/geoblacklight/geoblacklight/blob/bda30cffcfaef8b61d2fa4bd91c5268a67f87f15/solr/conf/solrconfig.xml#L120-L148

hackartisan:
to match aardvark field names

There are 2 instances of `layer_id` that I wasn't sure about. I propose that a new issue be opened for those if needed.
https://github.com/geoblacklight/geoblacklight/blob/8df46b9eee9a0f88118e864ab9222ccd613e95b0/app/views/catalog/_index_split_default.html.erb#L2

https://github.com/geoblacklight/geoblacklight/blob/8df46b9eee9a0f88118e864ab9222ccd613e95b0/spec/features/wmts_spec.rb#L25

closes #1155


ewlarson (approved):
Nice work! I think the outstanding layer_id uses can be left.

PR #1170, opened 2022-02-17
Support Faraday 1.0 & 2.0 at the same time.

### LINKED ISSUE #1168: Support multiple versions of Faraday
### Descriptive summary

Geoblacklight 4 should support Faraday 1 & 2, if possible, as there are several upstream libraries that have a variety of requirements.

tpendragon:
Closes #1168

To do this I had to copy over the redirect middleware locally. According
to the [faraday upgrade
guide](https://github.com/lostisland/faraday/blob/main/UPGRADING.md#faraday-middleware-deprecation)
somebody will have to extract that middleware to its own gem and support
it. I don't have a desire to be that person on a global scale, but for
now we can have it here.

This also adds a test to make sure that redirects work and adds a matrix
to CI to test for Faraday 1.0 & 2.0.


hackartisan (commented):
looks good to me! Is there a way to make the gh actions check names more descriptive? For example, clearly label ruby 2.7 as opposed to just having "2.7"? And for the Faraday addition, as well?

.github/workflows/ruby.yml
```diff
@@ -21,6 +21,10 @@ jobs:
     strategy:
       matrix:
         ruby: [2.7, '3.0']
+        faraday_version: ['']
+        include:
+          - ruby: 2.7
+            faraday_version: '~> 1.0'
```
eliotjordan, line 27: suggestion: Could we explicitly declare Faraday '~> 2.0' for ruby 3.0? It would make the config easier to read and understand.

eliotjordan (changes requested):

.github/workflows/ruby.yml
tpendragon: You mean instead of the `['']`? I considered that and was concerned that the faraday version default would then need to be changed here and in `Gemfile.extra`. Am I overthinking it?

tpendragon: Also this isn't making Faraday 1 for 2.7, it's adding a third matrix item which is ruby 2.7 and faraday 1.

eliotjordan:
@hackartisan Improving the CI config is a good idea, but I think it's out of scope on this PR. I opened a new issue for it: https://github.com/geoblacklight/geoblacklight/issues/1177

.github/workflows/ruby.yml
eliotjordan: I see. I was confused as well. I think this is fine.

eliotjordan (approved):

PR #1172, opened 2022-02-18
Always publish coverage data

cbeer:
Otherwise, if the build fails because the coverage is too low, the coverage step is skipped and the artifacts aren't published.


ewlarson (approved):
👍

PR #1180, opened 2022-02-18
Vendorized JS/CSS: update libraries/packages

ewlarson:
We vendorize Leaflet, because of inline image/css references that need an asset_path, and a few additional JS libraries used in GeoBlacklight. This PR updates them to their most recent releases. It also makes setting Leaflet's default icon directory path configurable in `settings.yml`, should anyone want to customize the icons they are using. 

* Update: leaflet (1.4.0 to 1.7.1)
* Update: esri-leaflet (2.2.1 to 3.0.4)
* Update: leaflet-iiif (2.0.1 to 3.0.0)
* Remove: leaflet-label.css (unnecessary post leaflet 1.0)


lib/generators/geoblacklight/templates/settings.yml
```diff
@@ -160,6 +160,7 @@ OPACITY_CONTROL: &opacity_control
 
 LEAFLET:
   MAP:
+  ICON_DEFAULT_IMAGEPATH: ".."
```
eliotjordan, line 163: question: Is this a parameter that we need to make configurable to users? I guess I'm wondering what would make cause a implementation to change this value.

ewlarson: Yeah, I was definitely under the influence of some stackoverflow comments where people were adopting different SVG iconography libraries. You might swap in a public dir path if you had a font awesome pro license and wanted their "duotone" look instead of Leaflet's default.

But, if you're willing to go that far, you're likely going to override GBL's viewer code... Maybe this is just unecessary? I'll think about it for a bit.

eliotjordan: If we had a use case / user request then I think we should consider it. I'm concerned that it's extra complexity in an already complex settings file.

ewlarson: Sure. I've rolled that change back. We still need to set the path ourselves, otherwise Leaflet+Rails asset_url defaults the icon path to '//assets' and icons do not render.

app/assets/javascripts/geoblacklight/viewers/viewer.js
```diff
@@ -10,6 +10,11 @@ GeoBlacklight.Viewer = L.Class.extend({
 
     L.Util.setOptions(this, options);
 
+    var leafletOptions = this.data.leafletOptions;
```
eliotjordan, line 13: issue: Unused variable. Should remove.

eliotjordan (changes requested):

eliotjordan (approved):

PR #1194, opened 2022-02-23
Combine download and export cards into dropdown button

### LINKED ISSUE #1131: Combine "Downloads" and "Export formats" content areas
### Descriptive summary

During usability testing at Princeton, testers were given pause by the separate content areas for "Downloads" and "Export formats". From the user perspective it doesn't seem that this distinction is important, and it creates a decision point that stalls their progress.

Recommend displaying the contents of "Export Formats" in the "Downloads" box and removing the separate content area.

If implemented we would like to see this on main and backported to the 3.x branch.

### Screenshot

Screen Shot 2022-02-08 at 10 34 32 AM

thanyabegum:
Closes #1131 by changing the Download and Export cards in the sidebar to a single Downloads dropdown button.

**Before**
Screen Shot 2022-02-23 at 2 17 52 PM

**After**
Screen Shot 2022-02-23 at 2 21 27 PM


tpendragon:
@thanyabegum Can you add a screenshot to this PR when you get a chance?

thanyabegum:
> @thanyabegum Can you add a screenshot to this PR when you get a chance?

Will do!

eliotjordan:
I really like combining the downloads into a single place. It looks great _except_ in places where there is  "Derived Datasets" / related content section. Instead of dropdown, I wonder if this should be a an accordion panel. I suppose the drop down could be positioned last, I just worry about it getting lost in the page.

karenmajewicz:
Have we considered moving the Related Items widgets to the bottom of the metadata? Here is how the Big Ten is doing it example: https://geo.btaa.org/catalog/675283de1b4c429387bccc06655f9cee_1 We might not want default GBL to look exactly like that, but it does free up real estate in the right hand column.

hackartisan:
@eliotjordan we could try it as an accordion. I see what you mean about the content overlapping. We might still want to style it like a button, to make it a matched pair with the web services button.

app/views/catalog/_downloads_dropdown.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <%= download_link_hgl(download_text(document.download_types.first[0]), document) %>
+<% end %>
+
+<% if document.iiif_download.present? %>
+  <%= download_link_iiif %>
+<% end %>
+
+<% if document.download_types.present? %>
+  <% document.download_types.each do |type| %>
+    <%= download_link_generated(type.first, document) %>
```
hackartisan, line 25: I think we want this to keep the translation behavior that was previously in the helper method `download_generated_body`

hackartisan:
Here are some screenshots of the accordion / collapse implementation: 
Screen Shot 2022-02-24 at 12 02 25 PM
Screen Shot 2022-02-24 at 12 02 31 PM

tpendragon (approved):

hackartisan (changes requested):
block on getting these translations back.

eliotjordan (approved):

hackartisan (approved):
Created https://github.com/geoblacklight/geoblacklight/issues/1200 to un-block merge of this ticket, since another 4.0 ticket depends on this getting merged.

PR #1207, opened 2022-02-24
Update item view fields on catalog_controller.rb

karenmajewicz:
This branch:

1. Cleans up the ITEM VIEW FIELDS. There were some duplicated lines in the commented-out code.
2. Turns on all of the descriptive/ string value fields by default
3. Re-orders the fields to match the Aardvark schema order.

Addresses #1052 In the previous version, we had been more conservative about only showing a few fields. However, I think it is better to display them all by default. This allows adopters to evaluate the fields and the examples shown in the fixtures. It would be easier for adopters to see extraneous fields and hide them rather than not know they existed.


karenmajewicz:
Thanks for fixing the tests, Eric. Is this seem like a good idea to display all the fields?

eliotjordan (approved):
This looks great! 👍

PR #1208, opened 2022-02-25
Move web services link out of show tools to a button

hackartisan:
Moves the web services link to give it visual symmetry with the download button, since they serve similar use cases (namely, using the data).

It also updates the web services modal to remove a trailing line separator.

### Upgrade notes
If you take no action (on an unmodified geoblacklight instance) you will have both web services links. To remove the link from the tool versions and keep the new button working:
* In your app's catalog controller
   * Remove the line beginning `config.add_show_tools_partial :web_services`
   * Add a small controller action to keep the web services modal working. This code was previously generated as part of blacklight's `add_show_tools_partial` functionality. The method you need can be copied from the updated catalog controller template and is called `web_services`.

#### Button
Screen Shot 2022-02-25 at 9 56 51 AM
Screen Shot 2022-02-25 at 9 57 06 AM

#### Modal before

Screen Shot 2022-02-25 at 9 29 02 AM

#### Modal after

Screen Shot 2022-02-25 at 9 28 00 AM

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/downloads.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/downloads.scss
@@ -1,4 +1,4 @@
-// Styling for Download dropdown button
+// Styling for Download and Web services buttons
 .sidebar-btn {
   width: 100%;
 
@@ -7,7 +7,8 @@
   }
 }
 
-.downloads-sidebar {
+/* match bottom margin on tools card */
+.downloads-sidebar, .web-services-sidebar {
   margin-bottom: 20px;
 }
 
--- a/app/assets/stylesheets/geoblacklight/modules/downloads.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/downloads.scss
@@ -8,8 +8,12 @@
 }
 
 /* match bottom margin on tools card */
-.downloads-sidebar, .web-services-sidebar {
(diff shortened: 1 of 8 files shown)


ewlarson (approved):
I like this look and it works well.

The only downside to this approach is `web_services` would no longer support Blacklight's catalog_controller config mechanics: ex. `config.add_show_tools_partial` or `config.show.document_actions.delete`

But... I can't imagine why someone implementing GBL wouldn't want these on by default anyway. I think this is good to ship until we revisit this as a view_component sometime in the future.

hackartisan:
@ewlarson that's interesting. The downloads also don't support that kind of configuration, right? Would we want that to, as well?

ewlarson:
I honestly vacillate between liking the catalog_controller config setup and thinking it's just wacky for a controller to do so much. My comment above is mostly a reminder to future "us" that we chose this path consciously.

I appreciate keeping things configurable as it provides the most flexibility for everyone, but I think we'll need to revisit a lot of our GBL html views in GBL v5 (BL v8) as view_components and that's the best opportunity to really rethink our feature implementations and how they fit in BL's config tooling.

I think we ship this as it stands. I vote to 🚢

hackartisan:
@kgjenkins In the before image there is an extra line at the bottom.

Issue #1177, opened 2022-02-18
Improve GitHub Actions configuration


eliotjordan:
### Descriptive summary

This is a ticket to request improvements to the GitHub Actions configuration file:
https://github.com/geoblacklight/geoblacklight/blob/a567fc2e5b2657c0d1a4d00b8ae47d29ed98e58c/.github/workflows/ruby.yml

Improvements:

1. Make the job titles more descriptive when using a matrix of Ruby versions. Right now, the titles say test 2.7 and test 2.6 and it isn't clear what that's in reference to.
2. Is there a way we can utilize [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) to streamline the configuration?

karenmajewicz:
Is there prior art at Princeton that we can adopt for this?

thatbudakguy:
this may be done? the workflow file was cleaned up so there's only one job, and we now test several more version combos that we didn't before. also, the job names accurately reflect what versions are being tested.

eliotjordan:
This ticket is done. The tests have more descriptive names now:

```
CI / test ruby 3.2 / rails 6.1.7.6
CI / test ruby 3.2 / rails 7.1.3
CI / test ruby 3.3 / rails 6.1.7.6 
CI / test ruby 3.3 / rails 7.1.3
```

PR #1214, opened 2022-02-25
Adjust GitHub actions matrix build

thatbudakguy:
* Collapse extra rails 6 job into include
* Specify bundler version in matrix
* Include matrix versions in job names
* Change rails versions in gemspec to match tested versions

See #1177

I'm a little unfamiliar with bundler dependency syntax, but I think this more accurately captures what we want to test/say we support.

- We allowed Rails 5 in the gemspec, but didn't test it. Testing it failed, so I don't think we should support it.
- There was some stuff related to testing Rails 7, but our required version of Blacklight doesn't support it, so we can't support it. I think it wasn't actually getting used because the environment variable wasn't named correctly (?)

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -24,93 +24,49 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [2.7, '3.0']
-        faraday_version: ['']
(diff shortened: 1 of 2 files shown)


thatbudakguy:
@cbeer fwiw i think your fixes in blacklight mean that we can test rails 5 again; i opened this prior to that work

PR #1222, opened 2022-03-30
Bug fix: allow all Aardvark relationship fields to browse suppressed records

### LINKED ISSUE #1221: GBL v4 - Expand "Browse all" relationship links to include additional Aardvark fields
### Descriptive summary

PR forthcoming...

This bug impacts GeoBlacklight v4/Aardvark releases to date. With Aardvark, we've expanded the number of fields that can be used to show parent/child relationships: Relation, Member Of, Is Part Of, Source, Is Version Of, Replaces, and Is Replaced By. 

### Expected behavior

When viewing a record that has a non-Source field relationship (ex. Is Part Of), the "Browse all X records..." links should return a list of records carrying that relationship.

Screen Shot 2022-03-29 at 9 33 38 AM

### Actual behavior

The browse link returns zero results:

Screen Shot 2022-03-29 at 9 34 44 AM

Because! Our SuppressedRecordsSearchBehavior concern is only checking for Source field suppressed records:
https://github.com/geoblacklight/geoblacklight/blob/6f7ddfb91938b507c8d4083495e324511abfcf38/app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb#L16

Instead we need our query to un-suppress any relationship-capable field when browsing.

### Related work

Aardvark Relationships - https://github.com/geoblacklight/geoblacklight/issues/1099

ewlarson:
We need to allow all the new Aardvark relationship fields to bypass our suppressed records filter. This PR adds additional non-Source field relationship fixtures (the Wabash Topo docs) and a feature test to ensure our "Browse all X" link is working correctly.

Fixes #1221

### DIFF
--- a/app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb
+++ b/app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb
@@ -12,8 +12,13 @@ module Geoblacklight
     # @param [Blacklight::Solr::Request]
     # @return [Blacklight::Solr::Request]
     def hide_suppressed_records(solr_params)
-      # Show child records if searching for a specific source parent
-      return unless blacklight_params.fetch(:f, {})[Settings.FIELDS.SOURCE.to_sym].nil?
+      # Show suppressed records when searching relationships
+      return if blacklight_params.fetch(:f,
+                                        {}).keys.any? do |field|
+                  Settings.RELATIONSHIPS_SHOWN.map do |_key, value|
+                    value.field
+                  end.include?(field)
+                end
 
       # Do not suppress action_documents method calls for individual documents
       # ex. CatalogController#web_services (exportable views)
--- a/config/locales/geoblacklight.en.yml
+++ b/config/locales/geoblacklight.en.yml
@@ -73,6 +73,7 @@ en:
       descendant: 'Derived Datasets'
       member_of: 'Belongs to collection...'
       part_of: 'Is part of...'
+      part_of_descendants: 'Has part...'
       relation: 'Related Records'
       replaces: 'Item replaces...'
       replaced_by: 'Item replaced by...'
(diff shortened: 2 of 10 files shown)


ewlarson:
Two tests fail here because https://map-tiles-staging.princeton.edu/mosaicjson/tiles/WebMercatorQuad is currently down.
Returns:
```
{"message":"Internal Server Error"}
```

@eliotjordan / Is there a better URL for these tests to ping?

app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb
```diff
@@ -12,8 +12,13 @@ module SuppressedRecordsSearchBehavior
     # @param [Blacklight::Solr::Request]
     # @return [Blacklight::Solr::Request]
     def hide_suppressed_records(solr_params)
-      # Show child records if searching for a specific source parent
-      return unless blacklight_params.fetch(:f, {})[Settings.FIELDS.SOURCE.to_sym].nil?
+      # Show suppressed records when searching relationships
+      return if blacklight_params.fetch(:f,
+                                        {}).keys.any? do |field|
```
eliotjordan, line 17: issue: this would be easier to read if line 17 was moved up to line 16.

eliotjordan (changes requested):

eliotjordan:
@ewlarson This looks good except for a small formatting issue.

app/models/concerns/geoblacklight/suppressed_records_search_behavior.rb
ewlarson: Ahhh! Yeah, I agree it looks odd, but this is how Rubocop autocorrected the indentation. 🤷

eliotjordan: Rubocop. smh.

eliotjordan (approved):

PR #1224, opened 2022-05-09
Blacklight 7.25 compatible filter changes

barmintor:
add Blacklight 7.25+ compatible changes to `Geoblacklight::BboxFilterField`
avoid deprecated behaviors

### DIFF
--- a/app/models/concerns/geoblacklight/bbox_filter_field.rb
+++ b/app/models/concerns/geoblacklight/bbox_filter_field.rb
@@ -2,6 +2,14 @@
 
 module Geoblacklight
   class BboxFilterField < Blacklight::SearchState::FilterField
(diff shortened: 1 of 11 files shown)


barmintor:
Some of these changes are not strictly necessary if https://github.com/projectblacklight/blacklight/pull/2715 is released, but they remove the use of a deprecated helper.

barmintor:
Added a commit to address some linter failures in specs, though they are outside the scope of the change.

ewlarson (approved):
Hi @barmintor - Thanks for looking out for GeoBlacklight here. Much appreciated!

ewlarson:
Just a quick note to say the changes in the PR need to be added to main and released in new v4 and v3 versions of GBL.

eliotjordan (approved):

Issue #1239, opened 2022-07-13
Consider replacing Rubocop with Standard


eliotjordan:
Could help to keep Rubocop update churn to a minimum.
Tag line: “No decisions to make. It just works.”

https://github.com/testdouble/standard

ewlarson:
Placing "On Hold" for future discussion.

dmolesUC:
On the current head (5b433bc), `standardrb` reports 2978 warnings on 39 separate cops, so unless you really, really like their opinionated style, I'm not sure this is such a good idea.

If RuboCop churn is the problem, that can be solved by setting `NewCops: disable` (currently `enable`) in `.rubocop.yml`.

eliotjordan:
I do like the opinionated style. 🤷‍♂️  Our rubocop config and rubocop-todo is bonkers.

eliotjordan:
Ran standardrb --fix. Not too bad really. I like it. https://github.com/geoblacklight/geoblacklight/pull/1244

karenmajewicz:
we will try this out after version 4 comes out

dmolesUC:
I'll be sad to see the single-quoted string literals go, but 🤷

eliotjordan:
@dmolesUC It is possible to set exceptions, but not having to worry about customizing the linter is compelling. If single-quoted strings are a community preference, we should talk about it. People have passionate opinions about this topic. 😄

ewlarson:
This change has been made. Using Standard now

PR #1244, opened 2022-08-17
Replace rubocop with standardrb

eliotjordan:
Example of switching to standardrb and running `--fix`. The handful of violations that weren't auto-fixed could be easily addressed. 

Refs #1239 

Personally, I think this is great. I like the highly opinionated defaults. Keeps us from making exception after exception as our rubocop config file grows.


ewlarson (approved):
This looks good. I've added a .standard.yml file here, with the format progress option so we can see the linter's output in our CI runs. Also, I tested that CI does indeed fail if you break a cop. This is ready to 🚢

PR #1250, opened 2022-09-14
GitHub Actions: Bump Ruby to 2.7

ewlarson:
Latest capybara release requires Ruby 2.7 minimum. Bumping our matrix requirement to match.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -25,18 +25,9 @@ jobs:
     strategy:
       matrix:
         ruby_version: ['3.0', 2.7]
-        rails_version: [6.1]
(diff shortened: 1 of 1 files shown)


thatbudakguy:
closes #1251

thatbudakguy:
i think we can actually just delete both of those special-case stanzas from the build, no? rails 7 supports ruby 2.7, so as long as we end up with:
```yaml
      matrix:
        ruby_version: ['3.0', 2.7]
        rails_version: [7.0.2.2, 6.1]
        bundler_version: [2.1.1]
        faraday_version: ['>= 2', '~> 1.0']
```
our bases are covered.

thatbudakguy (approved):

Issue #1274, opened 2023-03-01
Add support for a new 'display note' metadata field


rmseifried:
### Descriptive summary

The metadata group would like to add a new field `gbl_displayNote_sm` to the Aardvark metadata schema (for more discussion, see [OGM issue #16](https://github.com/OpenGeoMetadata/metadata-issues/issues/16)). Several institutions have been using custom fields for relevant information about access methods, data quality concerns, etc., and we would like to create a standardized field to contain this information. 

Decisions made:
- Property name: `gbl_displayNote_sm`
- Multivalued: yes
- Indexed: no
- Versioning: will be added to to the existing [Aardvark metadata schema](https://opengeometadata.org/docs/ogm-aardvark) (i.e. a new version will not be issued)

### Desired behavior

Eventually we would like the default application to display this text as a callout/admonition below the title, as in [this example from NYU](https://geo.nyu.edu/catalog/nyu-2451-38625). Ideally, there would be several options for how to style the callout, based on tags in the metadata. The tags we want currently:
- Warning
- Info
- Tip
- Note (default)

Example from [Docusaurus](https://docusaurus.io/docs/markdown-features/admonitions):
![Screenshot 2023-03-01 at 4 58 56 PM](https://user-images.githubusercontent.com/36506059/222274483-2445d336-6965-4c39-83e7-0efb7c6a342b.png)

### Feedback wanted: how to structure the field?

If the field were structured as below, would it be possible to code GBL so that it automatically detects the callout tag and chooses the applicable styling? In this example, there would be four separate callouts.

```
  "gbl_displayNote_sm": [
    "Warning: This text (starting with 'This text') will be displayed in a red box",
    "Info: This text (starting with 'This text') will be displayed in a blue box",
    "Tip: This text (starting with 'This text') will be displayed in a green box",
    "This is text without a tag and it will be assigned the default 'note' style"
  ]
```

Or is there a better/easier way to structure the field to achieve the same result?

ewlarson:
I can definitely see the value of this proposed new element. As it is currently imagined, it does create a bit of a technical challenge, albeit one we've dealt with in the past...

Our data store, Apache Solr, works best when a metadata field has one or more values (flat data):

```json
{ "dct_title_s": "Once upon a time." }
```

```json
{ 
    "dct_alternative_sm": [
        "Once upon a time." ,
        "Twice upon a time."
    ]
}
```

[...]

PR #1278, opened 2023-03-03
add displayNote test record

kgjenkins:
This adds a test record for the new Display Note field as described in #1274


ewlarson (approved):
👍

PR #1282, opened 2023-03-08
add `gbl_displayNote_sm` to validation file

rmseifried:
I also made this change in the JSON validation file stored on the OpenGeoMetadata site (https://github.com/OpenGeoMetadata/opengeometadata.github.io/tree/main/schema).

Related question: is there a reason why there are two copies of these validation files?

### DIFF
--- a/schema/geoblacklight-schema-aardvark.json
+++ b/schema/geoblacklight-schema-aardvark.json
@@ -24,6 +24,12 @@
         "type": "string"
       }
     },
+    "gbl_displayNote_sm": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
     "dct_creator_sm": {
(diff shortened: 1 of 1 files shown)


karenmajewicz (approved):

karenmajewicz:
At some point, a few of us thought it would be better to keep the schema files in OpenGeoMetadata. However, before deleting these, we asked around and some people prefer to have the schema files directly in the GeoBlacklight application. 

I think GeoCombine may have used these files at some point for validation, but I am not sure if that is still the case.

ewlarson:
Yes! Exactly what Karen said. I consider the "official" schema file the one in the OGM repo, but we need a copy here in GBL to sanity check our test fixtures and for local metadata validation steps, too.

PR #1300, opened 2023-06-14
Settings: prepend the new v4.1 settings.yml additions

ewlarson:
This will help keep the v4+ upgrade path as smooth as possible.

Also, this drops the old v3.4 settings file additions. We only need v4 updates here.


dl-maura (approved):
tests are passing in GitHub

Issue #1293, opened 2023-06-07
Create a link on item view pages to browse other records in the same "collection" or group


karenmajewicz:
### Descriptive summary

In OGM Aardvark, collections and groupings are defined by adding the ID of the parent record into one of these fields:

- [Member Of](https://opengeometadata.org/ogm-aardvark/#member-of)
- [Is Part Of](https://opengeometadata.org/ogm-aardvark/#is-part-of)

This ticket proposes to improve the user browsing experience by including a link from an item page directly to a list of siblings.

### Current user flow example

If a user lands on an item page and wants to see other datasets from the same collection, they need to make multiple actions.

1. User lands on the item page (https://geo.btaa.org/catalog/CB2D6C12-BFFF-4DC8-853A-EB512E93DC47)
2. User scrolls down and finds the "Is Part Of" widget
3. User clicks link and is taken to the Parent record (https://geo.btaa.org/catalog/10d-03)
4. User scrolls down and finds the "Has Part" widget
5. User clicks "Browse all xxx records..." and is taken to a list of records in the same collection (https://geo.btaa.org/?f%5Bdct_isPartOf_sm%5D%5B%5D=10d-03)

See this gif:

![wi-isPartOf](https://github.com/geoblacklight/geoblacklight/assets/2367677/590119ce-2ecc-4f43-be8f-814b357a4bab)

### Proposed user flow

We believe that users would like to be able to access the sibling records without having to visit the parent first.  

1. User lands on the item page (https://geo.btaa.org/catalog/CB2D6C12-BFFF-4DC8-853A-EB512E93DC47)
2. User finds the relationship widgets
3. User can choose between:
   - "See other items in this collection" (https://geo.btaa.org/?f%5Bdct_isPartOf_sm%5D%5B%5D=10d-03)
   - "Read about this collection" (https://geo.btaa.org/catalog/10d-03)

### Design

I am not sure what the actual design or layout should be yet. Some options:

- there could be a single widget box called "Collection" that has two links in it
- there could be two widget boxes  - one for a link to parent and one for a link to siblings
- this could be folded into the text of the metadata content itself rather than called out as part of a widget

dl-maura:
2 draft PRs

https://github.com/geoblacklight/geoblacklight/pull/1301

https://github.com/geoblacklight/geoblacklight/pull/1302

Two options we could look at, or a possible other way?

karenmajewicz:
Next steps:

[...]

PR #1302, opened 2023-06-15
Relationships - Browse link for parent descendants - Example 2

ewlarson:
Addresses #1293

This is an example of how a browse link for "collection" relationships could be added to GBL.

* Pros: Avoids rendering a browse link for 1:1 relationships, includes browse search result count
* Cons: Adds queries for inverse relationships into the view (an anti-pattern), data would be better captured in the relationship response/query class(es).

### Child Item / Links to Parent Browse

Screenshot 2023-06-15 at 10 32 26 AM

### Parent

Screenshot 2023-06-15 at 10 32 22 AM


thatbudakguy:
Now implemented by #1908.

Issue #926, opened 2020-07-08
Add a copy to clipboard action to the Citation modal


ggeisler:
An enhancement to the recently merged citation feature (#914), it might be nice for the user to be able to select a button to copy the citation, rather than having to manually select the citation text and copy it themselves. Especially since it seems like most of the time a user chooses to display the Citation modal, they will be doing so to copy the citation.

### Proposed enhancement

- [ ] Add a modal footer with "Copy citation" and "Close" buttons. I'd make "Copy citation" a primary button and "Close" a secondary.

<img width="929" alt="Screen Shot 2020-07-08 at 10 08 00 AM" src="https://user-images.githubusercontent.com/101482/86949908-3ca99c80-c104-11ea-8ea0-07694f6b2bf0.png">

---

- [ ] When the user selects the "Copy citation" button, the citation text should be copied to the user's clipboard.

- [ ] To make clear to the user the outcome of their action, display an alert with the message "Citation copied to your clipboard".

<img width="929" alt="Screen Shot 2020-07-08 at 10 06 49 AM" src="https://user-images.githubusercontent.com/101482/86950091-898d7300-c104-11ea-82c8-ae4f600aec9b.png">

eliotjordan:
@ggeisler Should this issue be moved upstream to Blacklight?

ggeisler:
@eliotjordan In theory, it does seem like a useful enhancement to the Blacklight cite tool. I filed the ticket here, however, because we implemented a more customized citation feature in GeoBlacklight (https://github.com/geoblacklight/geoblacklight/pull/914) and I saw this as something I should've included in the design for that originally.

As a UX designer, I think I'm not really qualified to judge the best repo to implement the solution. @mejackreed might have a better sense of where it should go.

(Also, based on how I see the current cite tool working in Blacklight -- it only shows the title of the item as the citation -- it seems sort of dysfunctional so it feels weird to add a copy feature to it, but that might be irrelevant to the decision about where to implement it.)

ewlarson:
See #1209 too

ewlarson:
Closing via https://github.com/geoblacklight/geoblacklight/pull/1307/

PR #1307, opened 2023-06-29
Add Citation to Clipboard

the-codetrane:
Pull Request for https://github.com/geoblacklight/geoblacklight/issues/926

CitationModal

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ jetty
 yarn-debug.log*
 .yarn-integrity
 yarn.lock
+
+# Ignore IDE files
+.idea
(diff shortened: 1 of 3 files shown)


ewlarson (changes requested):
@the-codetrane // Thanks so much for taking this feature on and my apologies for the long wait on a review (peak summertime).

These changes look good to me, but lets do a couple things before merging this into GBL:
- [x] Remove the comment above the script tag, I think the js code placement is fine
- [x] Add a spec/feature test just to illustrate the new copy button appears when clicking "Cite"

You might look at the spec/features/metadata_panel_spec.rb for something similar. We don't have much javascript-based testing in the application at all. But just a simple DOM check that the button renders would be nice.

Thanks again!

ewlarson (approved):
👏

PR #1336, opened 2024-02-02
Use Clover for IIIF Manifests

eliotjordan:
- Clover: vanilla javascript approach
- Use Clover IIIF with configuration


ewlarson (approved):
This is great! 🚀

PR #1333, opened 2024-02-01
add baseline support for OpenIndexMaps Specification 1.0.0

srappel:
adds an OpenIndexMaps example that uses the newest version of the [OpenIndexMaps specification (v1.0.0)](https://openindexmaps.org/specification/1.0.0)

PR #1339, opened 2024-02-06
Moves OpenIndexMap test fixtures to repo

srappel:
This creates a new directory in the fixtures folder called index_maps and populates it with 5 geojson files. This is a companion PR to #1333 which will (among other things) redirect the solr_documents for index maps to point to these instead of pointing to various github repo external to the geoblacklight organization.


karenmajewicz (approved):
Looks good. I will merge, since these are just fixtures.

PR #1347, opened 2024-02-09
Updates to test app generation

eliotjordan:
- Symlink fixture directories: allow dev to update fixtures without regenerating test app.
- Remove unneeded sqlite fix

Some improvements to test app generation. Will help with exploring development tools like Lando.


ewlarson (approved):
👍

PR #1348, opened 2024-02-09
update readme to point to external docs #1345

### LINKED ISSUE #1345: README: point to geoblacklight.org for install docs

mradamcox:
Moves installation info out from the README and points to the external documentation on the geoblacklight website. Our main goal with this change to avoid having the same content/instructions in two different places (here and also in the external documentation).

Closes #1345


karenmajewicz (approved):
Looks good - we will need to add the geoblacklight.org/docs page to the website.

Issue #1352, opened 2024-03-01
Vite/JS - Cannot load local vite entrypoint in application


ewlarson:
### Descriptive summary

The latest GBL v4.2 release introduced [Vite Ruby](https://vite-ruby.netlify.app/) for frontend javascript management within our GBL Rails engine. Vite is used to load OpenLayers and our OpenLayer-based COGs and PMTiles map viewers: https://github.com/geoblacklight/geoblacklight/pull/1332

However, the current implementation of Vite within GeoBlacklight doesn't allow someone to add additional Vite entrypoints into their local application.

### Example

Add a `local.js` vite entrypoint to the local Rails application:

![Screenshot 2024-03-01 at 12 00 05 PM](https://github.com/geoblacklight/geoblacklight/assets/69827/f2597111-3fc9-47e9-9e7d-fe58bb8625a4)

Load it in `base.html.erb`

![Screenshot 2024-03-01 at 12 02 30 PM](https://github.com/geoblacklight/geoblacklight/assets/69827/000fe092-4675-46a0-9f35-76b5cdc3a849)

See it fail...

![Screenshot 2024-03-01 at 11 55 13 AM](https://github.com/geoblacklight/geoblacklight/assets/69827/7c4dc232-3a9b-4243-81e5-cfcb68e7055e)

### Expected behavior

We should allow (or recommend even), that GBL adopters use Vite Ruby as their local javascript management tool. Also, our current Vite implementation makes it difficult for additional GBL Rails engines, like GeoBlacklight Admin, to adopt and use Vite.

### Actual behavior

The new `local.js` file is not present/included in the Vite manifest.

### Steps to reproduce the behavior

Documented above.

### Related work

[Vite PR](https://github.com/geoblacklight/geoblacklight/pull/1330)

dl-maura:
@eliotjordan @ewlarson is this still relevant? We thought this was fixed in the latest release

ewlarson:
Closing via the v4.4 release.

PR #1353, opened 2024-03-04
Vite: uses local application entrypoints with an aliased path for importing engine code

ewlarson:
Uses vite.config.ts resolve[:alias] to provide path to Rails engine code.

Addresses #1352

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -65,7 +65,7 @@ jobs:
       run: bundle exec rake ci
       env:
         RAILS_VERSION: ${{ matrix.rails_version }}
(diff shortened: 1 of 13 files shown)


ewlarson:
Heya @eliotjordan

I'm seeing a couple wrinkles implementing Vite Ruby for frontend management within the BTAA Geoportal (or any GBL install). The gist of the trouble is documented here: https://github.com/geoblacklight/geoblacklight/issues/1352 

This PR is my first stab at making Vite work better for the local application, while also pulling code from inside the GBL Engine. Y'all at Princeton definitely have more Vite experience than I do. Have you all run into issues using Vite within local and Rails Engine-based applications?

eliotjordan (approved):

PR #1355, opened 2024-03-22
Vite/JS - Updates to run successfully in EngineCart and from template.rb

ewlarson:
Working to resolve adoption issues with the new Vite/JS features in GBL v4.2 and v4.3 releases.

These changes appear to resolve issues with running Vite/JS inside of EngineCart's .internal_test_app and via a GBL application built by our template.rb script.

The crux here is that pulling JS dependencies and dependency/import paths out of a Rails Engine and into the top-level Rails application is hard.

JS features/functionality would be best packaged into a NPM release for consuming at that top-level application — this is how Blacklight does it and also seems to be the best practice from Webpacker, too:
https://github.com/rails/webpacker/issues/21

### DIFF
--- a/app/frontend/stylesheets/openlayers.css
+++ b/app/frontend/stylesheets/openlayers.css
@@ -1,1 +1,0 @@
-// @import "../../../node_modules/ol/ol.css";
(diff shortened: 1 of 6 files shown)


ewlarson:
GBL v4.4 obviates this PR.

PR #1359, opened 2024-04-24
Add convertToRGB to prevent color errors

michaelshensky:


michaelshensky (commented):
By setting converToRGB to true here, it will fix an issue with COGs that utilize JPEG compression which causes the color model to be interpreted as YCbCr instead of RGB. Other applications like QGIS interpret the COG as RGB but Open Layers will not do so by default and will instead display a color inverted looking version of the map if convertToRGB is not used.

michaelshensky (commented):
> By setting convertToRGB to true here, it will fix an issue with COGs that utilize JPEG compression which causes the color model to be interpreted as YCbCr instead of RGB. Other applications like QGIS interpret the COG as RGB and display the colors correctly by default but Open Layers will not do so by default and will instead display a color inverted looking version of the COG if convertToRGB is not used.

marlo-longley:
I am not super familiar with openlayers and GeoTIFF, but this seems reasonable and aligns with the documentation (https://openlayers.org/en/latest/apidoc/module-ol_source_GeoTIFF-GeoTIFFSource.html) here.

Is there a specific example or testing path we could use to verify the colors display as intended?

marlo-longley (approved):

PR #1374, opened 2024-06-25
move home and search result maps to location_leaflet_map_component

### LINKED ISSUE #1271: ViewComponent - Homepage - Map
Add a VC for the homepage map. The feature should provide a mechanism for setting the default map extent and zoom level. It might also allow for configurable Leaflet plugins (geosearch, layer previews, clustering engine, etc.).

Screenshot 2023-02-28 at 5 08 05 PM

### LINKED ISSUE #1273: ViewComponent - Index - Map
Add a VC for the search results map. We should consider making the map "sticky" to the top of the window so results far down the page will still display their rollover bbox on the map (both BTAA and HGL are doing this now).

Screenshot 2023-03-01 at 1 00 58 PM

### Stanford (GBL default / Map is NOT sticky)
Screenshot 2023-03-01 at 1 21 03 PM

### BTAA Geoportal
Screenshot 2023-03-01 at 1 03 38 PM

### HGL 
Screenshot 2023-03-01 at 1 04 35 PM

dnoneill:
closes #1271  and closes #1273

### DIFF
--- a/app/components/geoblacklight/location_leaflet_map_component.rb
+++ b/app/components/geoblacklight/location_leaflet_map_component.rb
@@ -2,10 +2,11 @@
 
 module Geoblacklight
   class LocationLeafletMapComponent < ViewComponent::Base
-    def initialize(id: "map", label: nil, data_map: "", map_geometry: nil)
+    def initialize(id: "map", label: nil, data_map: "", map_geometry: nil, classes: "")
       @id = id
       @label = label
       @data_map = data_map
+      @classes = classes
       @map_geometry = map_geometry || Settings.HOMEPAGE_MAP_GEOM
       super
     end
--- a/app/components/geoblacklight/location_leaflet_map_component.rb
+++ b/app/components/geoblacklight/location_leaflet_map_component.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class LocationLeafletMapComponent < ViewComponent::Base
+    def initialize(id: "map", label: nil, data_map: "", map_geometry: nil)
+      @id = id
+      @label = label
+      @data_map = data_map
+      @map_geometry = map_geometry || Settings.HOMEPAGE_MAP_GEOM
+      super
+    end
+
+    def before_render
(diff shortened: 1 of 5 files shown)


app/helpers/blacklight_helper.rb
```diff
@@ -6,6 +6,6 @@ module BlacklightHelper
   include Blacklight::CatalogHelperBehavior
 
   def render_document_sidebar_partial(document = @document)
-    super(document) + (render "relations_container", document: document)
+    super + (render "relations_container", document: document)
```
dnoneill, line 9: This got added because it kept breaking the linter.

app/components/geoblacklight/location_leaflet_map_component.rb
```diff
@@ -0,0 +1,17 @@
...
+  class LocationLeafletMapComponent < ViewComponent::Base
+    def initialize(id: "map", label: nil, data_map: "", map_geometry: nil)
+      @id = id
+      @label = label
+      @data_map = data_map
+      @map_geometry = map_geometry || Settings.HOMEPAGE_MAP_GEOM
+      super
+    end
+
+    def before_render
```
marlo-longley, line 13: If you are hardcoding `id: "map"` does it need to be in the init arguments?
Same with label -- if label is always nil and then overwritten during `before_render` to pull from the template file, does it need to be in the arguments?

dnoneill: That is definitely a possibility since for the two tickets I have used label and id are the same. They seemed like fields that would be ones you might want to have a different label for on different pages and the id is one I could see needing to be changable due to how leaflet works. If you wanted to put multiple leaflets on the page, being able to change those would be preferable. I can take it out for now, but I thought it would probably be good to no longer hardcode these fields.

marlo-longley: Sounds good, just curious!

hudajkhan (approved):
The map is now sticky too! Yay!

hudajkhan:
I tried this branch out locally and the search page map rendered properly AND it was sticky : )

PR #1376, opened 2024-06-26
move download panel into its own component

### LINKED ISSUE #1365: ViewComponent - Show - Downloads
Appears in the sidebar and offers options depending on the type of the item and its references.

Image

peetucket:
Fixes #1365 - move downloads panel to view component

~~HOLD i am now moving some other related bits into a component too~~


app/helpers/blacklight_helper.rb
```diff
@@ -6,6 +6,6 @@ module BlacklightHelper
   include Blacklight::CatalogHelperBehavior
 
   def render_document_sidebar_partial(document = @document)
-    super(document) + (render "relations_container", document: document)
+    super + (render "relations_container", document: document)
```
peetucket, line 9: standardrb auto correct did this?

app/views/catalog/_show_sidebar.html.erb
```diff
@@ -3,7 +3,8 @@
 
 <div class="sidebar-buttons">
 <%= render :partial => "show_web_services" %>
-<%= render :partial => "show_downloads" %>
+<%= render(Geoblacklight::DownloadLinksComponent.new(document:)) if document_downloadable? %>
+<%= render(Geoblacklight::LoginLinkComponent.new(document:)) %>
```
peetucket, line 7: logic in the component itself will decide if we actually need the login link to render or not

app/views/catalog/_show_sidebar.html.erb
```diff
@@ -3,7 +3,8 @@
 
 <div class="sidebar-buttons">
 <%= render :partial => "show_web_services" %>
-<%= render :partial => "show_downloads" %>
+<%= render(Geoblacklight::DownloadLinksComponent.new(document:)) if document_downloadable? %>
```
peetucket, line 6: why is `document_downloadable?` here instead of in the `render?` method in the component?  because it is defined in `GeoblacklightHelper` helper and includes a call to `user_signed_in?`, a helper method likely in Devise or elsewhere?  

which was causing me grief in testing -- i will look some more

app/views/catalog/_show_sidebar.html.erb
```diff
@@ -3,7 +3,8 @@
 
 <div class="sidebar-buttons">
 <%= render :partial => "show_web_services" %>
-<%= render :partial => "show_downloads" %>
+<%= render(Geoblacklight::DownloadLinksComponent.new(document:, downloadable: document_downloadable?)) %>
+<%= render(Geoblacklight::LoginLinkComponent.new(document:)) %>
```
peetucket, line 7: logic in the components will decide if we actually need the login link to render or not -- pass in the knowledge needed (`document_downloadable?` is defined in the `GeoblacklightHelper` module)

hudajkhan:
I tried this out locally and the download panel seemed to render correctly, both when the dataset is public and when it is restricted and needs to show a login and download button underneath.

hudajkhan (approved):

peetucket:
thanks for verifying the various states it could be in (wasn't sure how to trigger them all locally, and wasn't 100% confident the specs covered them all)

PR #1377, opened 2024-06-27
Extract search results into its own view component

### LINKED ISSUE #1272: ViewComponent - Index - Search Results
Add a VC for GBL search results. Our current design is quite different than Blacklight's default metadata definition list, which makes this feature a good candidate for drafting our own GBL component.

We can configure our GBL component in the `catalog_controller.rb` file like so:

```ruby
    # GeoBlacklight Defaults
    # 
    config.index.document_component = Geoblacklight::SearchResultComponent
```

We'd then more or less copy Blacklight's [DocumentComponent](https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/document_component.rb) into our application and update the class/view code to our needs.

marlo-longley:
Closes #1272

Following the instructions in the ticket, this is named `search_result_component`. 
Everything upstream in Blacklight is named `document_component`. Slightly confusing but I did leave comments.

Screenshot 2024-06-27 at 13 40 13


app/views/catalog/_document_split.html.erb
```diff
@@ -1,6 +1,8 @@
 <div class='row'>
   <div id="documents" class="documents-list col-md-6">
-    <%= render documents, :as => :document %>
+    <% documents.each_with_index do |document, index| %>
+      <%= render Geoblacklight::SearchResultComponent.new(document: document, document_counter: index) %>
```
dnoneill, line 4: Could you do with_collection instead of a loop? I am not 100% sure this work

<%= render Geoblacklight::SearchResultComponent.with_collection(documents.each_with_index) %>

marlo-longley: Somehow it automagically works to do 
`<%= render Geoblacklight::SearchResultComponent.with_collection(documents) %>
`
-- and it knows what to pass for the `document:` and `index:` params in the initializer in the VC Class.

hudajkhan:
Tried out the code and it works as expected.

hudajkhan (approved):

PR #1383, opened 2024-06-28
move citation to component

### LINKED ISSUE #1369: ViewComponent - Show Tools - Citation

dnoneill:
- Also fix the copy div to not copy the whole modal and just the citation.

closes #1369


app/models/concerns/geoblacklight/solr_document/citation.rb
```diff
@@ -1,30 +0,0 @@
-# frozen_string_literal: true
-
-module Geoblacklight
-  module SolrDocument
-    module Citation
```
jcoyne, line 5: removing this constant (and it's methods) is a breaking change.  Do we need to do that?

spec/components/geoblacklight/citation_component_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe Geoblacklight::CitationComponent, type: :component do
```
jcoyne, line 5:
```suggestion
RSpec.describe Geoblacklight::CitationComponent, type: :component do
```

Don't depend on monkeypatching.

app/models/concerns/geoblacklight/solr_document/citation.rb
dnoneill: No, I can add it back in. I am not sure why it is breaking change though.

marlo-longley (approved):

PR #1384, opened 2024-06-28
move the arcgis link to its own component

### LINKED ISSUE #1371: ViewComponent - Show Tools - Esri/ArcGIS

peetucket:
Fixes #1371 - move the arcgis link to a component

Also get rid of the associated helper and move the only method to the component (it's not used elsewhere).  Move the specs for it to the new component spec.


marlo-longley:
Hi @peetucket -- looks good! Thoughts -- 

1. Question, is it expected that the link doesn't properly open in ArcGIS? Do you have example links you were working with that you could share? Maybe it is a data problem and not a code problem, but it would be good to witness a working link before merging. This is what happened when I followed the link on http://localhost:3000/catalog/nyu-test-soil-survey-map from your branch:

Screenshot 2024-07-01 at 11 51 19

2. Other show tools that provide links to external websites use `target="_blank"` in order to open a new tab/window. [See the carto link here](https://github.com/geoblacklight/geoblacklight/blob/0d0bd5a59a6146cd2014b50eb4bb56bd282cfdc2/app/components/geoblacklight/carto_component.html.erb#L1) for example. This tool does not follow that pattern (I understand it was originally not using `target="_blank"` and that this is parity) -- but do we want to standardize that pattern? @dbranchini? 

3. This needs a rebase with the catalog_controller.

app/components/geoblacklight/arc_gis_component.rb
```diff
@@ -0,0 +1,25 @@
...
+    attr_reader :document
+
+    def initialize(document:, action:, **)
+      @document = document
+      super
+    end
+
+    # Generates an ArcGIS.com viewer url with params that can open content directly
+    # See: https://doc.arcgis.com/en/arcgis-online/reference/use-url-parameters.htm
+    def arcgis_link(arcgis_urls)
```
dnoneill, line 14: I don't think this is a necessary change but you could update this to just have the argis_urls to be document.arcgis_urls

peetucket: right, yes -- i just removed the need to pass in the urls and just pulled them straight from the document which is available to the component class (and updated the specs)

app/components/geoblacklight/arc_gis_component.html.erb
thatbudakguy, line 1: extremely nitpicky, but could we name these things `arcgis_foo` instead of `arc_gis_foo`? That was the convention for the `arcgis_helper` and it feels more natural to me for some reason.

thatbudakguy (approved):

app/components/geoblacklight/arc_gis_component.html.erb
peetucket: makes sense -- renamed ... will merge when green

Issue #1367, opened 2024-06-12
ViewComponent - Show - Relations


thatbudakguy:
Renders in the sidebar; all of the different [relations fields](https://opengeometadata.org/relations-fields/) can create boxes here.

![Image](https://github.com/geoblacklight/geoblacklight/assets/4924494/15fae27e-17d6-4c75-bf8d-4ae25c49bc1a)

dnoneill:
closed with https://github.com/geoblacklight/geoblacklight/pull/1385

PR #1385, opened 2024-07-01
Componentize relations

jmartin-sul:
Part of #1367  -- I think to close that ticket, assuming this PR is moving in the right direction, I'd also have wanted to
* change `app/views/catalog/_relations_container.html.erb` so that it renders a component and emits a deprecation warning
* change `app/helpers/blacklight_helper.rb`'s `BlacklightHelper.render_document_sidebar_partial` to render the same component as the deprecated view and to append that to the call to `super`, instead of rendering the deprecated view.
* I was also uncertain whether `app/views/relation/index.html.erb` and `app/views/relation/index.json.jbuilder` needed to be componentized, so that might be another thing missing from this PR.

I might be able to get to that in the morning -- we'll see.  But this was as far as I got on Friday, and I'm unlikely to be able to give much attention to this past tomorrow (Monday) morning, so I wanted to put up what I had so far.

AFAICT, I didn't need to alter existing tests, or write new ones, because this functionality is already covered by `spec/features/relations_spec.rb`, and there's no expected change in behavior with this work.

### has no relations
Screenshot 2024-06-30 at 5 05 34 PM

### has two types of relation
Screenshot 2024-06-30 at 5 05 50 PM

### has one type of relation
Screenshot 2024-06-30 at 5 05 59 PM

Also: just a note that as someone new to this project, i love that some of the examples in the documentation also exist in the fixture data, and are easily findable.  made testing much easier.


app/components/geoblacklight/relations_component.rb
jmartin-sul, line 1: thinking about this a bit more, `DisplayRelationsForTypeComponent` might have been better and more descriptive, since this component only displays relations for one type of relation.  may try to touch that up in the morning, if i have a chance.

thatbudakguy (approved):

PR #1386, opened 2024-07-02
remove duplicate university icons

### LINKED ISSUE #1342: De-duplicate icons
There are several duplicate icons with different labels in https://github.com/geoblacklight/geoblacklight/tree/122f7131992971ab3b2ef6e288646d6cb455adc2/app/assets/images/blacklight.

ex: https://github.com/geoblacklight/geoblacklight/blob/122f7131992971ab3b2ef6e288646d6cb455adc2/app/assets/images/blacklight/michigan-state.svg

and https://github.com/geoblacklight/geoblacklight/blob/122f7131992971ab3b2ef6e288646d6cb455adc2/app/assets/images/blacklight/michigan-state-university.svg

We should remove the older version of the icons with deprecated names.

dnoneill:
closes #1342 

Please note I updated app/assets/images/blacklight/geoblacklight-icons.json and config/locales/geoblacklight.en.yml as well because my assumption is without the icon we wouldn't want the filename in those files.


config/locales/geoblacklight.en.yml
karenmajewicz, line 1: I missed one when making these changes: indiana.svg should be indiana-university.svg

karenmajewicz, line 1: I am not quite sure how this file works. Do we need to add a mapping here for the icons that replaced the originals?  Etc. 

remove: ~~chicago: University of Chicago~~
add: university-of-chicago: University of Chicago

dnoneill: I don't believe so because university-of-illinois-urbana-champaign doesn't show up in that file but it shows up correctly in the app.

dnoneill: I believe this is outside the scope of this ticket. We should create a new ticket for this.

PR #1388, opened 2024-07-02
Moving attribute table into own component

### LINKED ISSUE #1366: ViewComponent - Show - Attribute Table
Displays below the map and changes to look up attribute information when the map is clicked, if a WFS/WCS link is available.

Image

In the Stanford embed viewer, we do this as a custom leaflet overlay, so it's inside the viewer instead of outside it:

Image

hudajkhan:
Closes #1366 .
What this pull request does:

- Create attribute table component
- Use the component html erb file to display the attribute table container
- Passes in the value from the  "show_attribute_table" geoblacklight_helper method to determine if anything should be rendered
- Adds a simple test


app/components/geoblacklight/attribute_table_component.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  ##
+  # A component for rendering attribute table
+  #
+  class AttributeTableComponent < ViewComponent::Base
+    attr_reader :show_attribute_table
```
dnoneill, line 8: Why do you have this if you aren't using it?

hudajkhan: You are correct : ) will remove

hudajkhan: Have removed

app/views/catalog/_show_default_attribute_table.html.erb
```diff
@@ -1,19 +1 @@
...
-        <tbody class='attribute-table-body'>
-          <tr>
-            <td class='default-text' colspan='2'><em><%= t('geoblacklight.show.click_map') %></em></td>
-          </tr>
-        </tbody>
-      </table>
-    </div>
-  </div>
-<% end %>
+<%= render Geoblacklight::AttributeTableComponent.new(show_attribute_table:show_attribute_table?) %>
```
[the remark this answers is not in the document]
hudajkhan: Changed it as suggested above.

app/views/catalog/_show_default_attribute_table.html.erb
```diff
@@ -1,19 +1 @@
-<% if show_attribute_table? %>
```
thatbudakguy, line 1: do we want to preserve this check? If so, I think all you need to do is add a `render?` to `AttributeTableComponent` that calls it

thatbudakguy (approved):

PR #1390, opened 2024-07-03
add iiif drag and drop icon

dnoneill:
Not sure where to update the documentation for this so I am going to do so in the PR.

1. By default this is enabled. I believe that we should be indicating a IIIF image with the drag and drop functionality by default as is standard with a lot of IIIF services.
2. To disable this functionality the user needs to remove or set `IIIF_DRAG_DROP_LINK` to blank.
3. By default the drag and drop link goes to the manifest url. If the user wants to update the URL they can with the `IIIF_DRAG_DROP_LINK`, for example if the want the link to go to mirador they will update the `IIIF_DRAG_DROP_LINK` to `https://projectmirador.org/embed?manifest=@manifest`. The `@manifest` will be replaced with whatever the page's manifest is.
4. You do need to include a parameter in the url for manifest or drag and drop doesn't seem to work, i.e. `?manifest=`

Screenshot 2024-07-03 at 12 14 57 PM


thatbudakguy:
Is it normal that the link could go to a viewer and embed the object, instead of being a manifest URL? I thought it was always supposed to be the manifest URL, otherwise "drag and drop" wouldn't work – if you drag a mirador link onto a viewer, nothing will happen, right?

dnoneill:
I am pretty sure as long as there is a ?manifest= in the url it can find it. It was hard to find docs on this but I modeled what we are doing here: https://github.com/sul-dlss/exhibits/blob/70b043be02fd09dafeeb1a9033de429063245358/app/helpers/application_helper.rb#L21-L25

thatbudakguy (approved):

PR #1344, opened 2024-02-09
WIP: Blacklight v8

ewlarson:
New BL8 branch with latest GBL contributions and ES6 javascript.

### Todo

- [ ] Bootstrap v5 migration

### Test Suite errors

- [ ] Show Tools > SMS form is empty
- [ ] geoblacklight_present() > Spec failing with undefined `action_name`

Issue #1132, opened 2022-02-08
Update / upgrade javascript


eliotjordan:
### Descriptive summary

GeoBlacklight javascript code could use modernization. We should have a discussion about this and write up a proposal for how we might upgrade. Since the javascript modules are modified in individual implementations, this could be a disruptive change and we should take that into account.

Possible topics to discuss:

1. Use of `let/const` vs. `var`.
1. Use of [leaflet class structure](https://github.com/geoblacklight/geoblacklight/blob/main/app/assets/javascripts/geoblacklight/geoblacklight.js#L33). 
1. Should we convert to ES6? Could address the class and module problem above.
1. **Blacklight will stop requiring JQuery**
1. BL has dropped support for IE 11
1. How to bundle and ship JS (BL Community question)

### Related work
https://github.com/geoblacklight/geoblacklight/issues/774

thatbudakguy:
thanks for opening this @eliotjordan! i think another thing to bring into this discussion is codifying what browsers the app should support (if that doesn't already exist somewhere). it looks like we are already using babel's `preset-env`, which provides a nice [browserslist integration](https://babeljs.io/docs/en/babel-preset-env#browserslist-integration). that means we could decide formally on what we want to support, create [a `.browserslistrc` or similar](https://github.com/browserslist/browserslist), and let babel do the work of including polyfills for the features we want to support.

i think there are lots of good arguments for converting to ES6, but i'm also not familiar with the landscape of individual implementations. that said, i think it's fair to expect disruptive changes as part of v4, so hopefully it wouldn't be too big of a surprise.

eliotjordan:
At the 2/2/22 Sprint Meeting, we decided to:

- Remove jQuery dependency
- Rewrite javascript as ES6
- Improve javascript coverage with Jest
- Further study and discuss bundling options

eliotjordan:
We've been experimenting with using [vite](https://vite-ruby.netlify.app/) and [jsbundling-rails (with esbuild)](https://github.com/rails/jsbundling-rails) to bundle javascript. Vite is seems very nice.

ewlarson:
IIIF Image/Manifest viewing is something that should be considered during this update/upgrade process:
https://github.com/geoblacklight/geoblacklight/issues/1279

karenmajewicz:
[...]

PR #1392, opened 2024-07-08
Modernize the javascript and bundle everything with Vite

thatbudakguy:
Ref #1132 

**You may need to completely delete `.internal_test_app` for the generators to re-run correctly.**

This updates all of the javascript to ES6 (borrowing somewhat from #1344). It also removes sprockets imports from these files and instead uses Vite to bundle all of the JS and CSS for the app.

The `app/assets/javascripts` directory goes away, in favor of everything being listed in `app/javascript` like the other entrypoints. Many structures that made use of jQuery have been converted into Stimulus controllers and are now Turbo-compatible.

In general, everything that was vendorized or referenced from a gem is now installed via npm and referenced from node_modules. This is the pattern blacklight uses to ship its JS and SCSS sources. This means that the generated app can ultimately choose its own bundling strategy, since all the source files are available.


app/javascript/geoblacklight/downloaders/downloader.js
```diff
@@ -0,0 +1,62 @@
+export default class GeoBlacklightDownloader {
+  constructor(el, options) {
+    this.el = el;
```
dnoneill, line 3: Does this need to be the same? I am guessing not because it seems to work locally but I want to put a note just in case.

https://github.com/geoblacklight/geoblacklight/blob/007b52cef878e7102ea637f1ae5b995a114abf02/app/assets/javascripts/geoblacklight/downloaders/downloader.js#L11

app/javascript/geoblacklight/downloaders/downloader.js
```diff
@@ -0,0 +1,62 @@
...
+    this.options = options || {};
+    this.options.spinner = document.createElement("div");
+    this.options.spinner.className =
+      "spinner-border spinner-border-sm float-right";
+    this.options.spinner.role = "status";
+    this.options.spinner.innerHTML = '<span class="sr-only">Downloading</span>';
+    this.options.spinner.style.display = "none";
+    // @TODO: This spinner doesn't appear to work in GBL Proper anymore.
+    //document.querySelector('.exports .card-header').appendChild(this.options.spinner);
```
dnoneill, line 12: Pretty sure we can get rid of all the spinner stuff. Unless downloader is being used somewhere else, right now there is no element in the code that uses the class exports. When a download is successfully rendered it creates a flash message and there is no modal.

app/javascript/geoblacklight/downloaders/downloader.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return response.json();
+      })
+      .then((data) => this.complete(data))
+      .catch((error) => this.error(error));
+  }
+
+  complete(data) {
+    this.downloading = false;
+    this.el.disabled = false;
```
dnoneill, line 42: This is not in the old code but it should be, download should have a line that says this.el.disabled = true.

dnoneill: Actually after investigations since these are links, disabled doesn't do anything. We need to create a disabled class for links for this to work.

app/javascript/geoblacklight/downloaders/hgl_downloader.js
```diff
@@ -0,0 +1,41 @@
...
+      },
+      body: JSON.stringify({ email: email })
+    })
+    .then(response => {
+      if (!response.ok) {
+        throw new Error('Network response was not ok');
+      }
+      return response.json();
+    })
+    .then(data => this.complete(data))
```
dnoneill, line 38: This doesn't do anything after you fill out the form. Pretty sure it is the same in the current version, just making a note.

app/javascript/geoblacklight/metadata_download.js
```diff
@@ -0,0 +1,48 @@
...
+    if (!refUrl) {
+      this.download.style.display = "none";
+    } else {
+      this.download.style.display = "";
+      this.download.setAttribute("href", refUrl);
+    }
+  }
+}
+
+export default function initializeMetadataDownload() {
```
dnoneill, line 30: Metadata download is working locally. I am guessing it has something to do with the fact this function used to be in modules and had a wrapper Blacklight.onLoad

https://github.com/geoblacklight/geoblacklight/blob/24aba2e82f4b386e0301d8beebf969727bcf7fb0/app/assets/javascripts/geoblacklight/modules/metadata.js#L3

app/javascript/geoblacklight/truncation.js
```diff
@@ -0,0 +1,28 @@
+// Truncate and expand/collapse abstracts (descriptions) on show page
```
dnoneill, line 1: This isn't working. Looking into it now.

app/javascript/geoblacklight/truncation.js
```diff
@@ -0,0 +1,28 @@
...
+    control.setAttribute("aria-expanded", "false");
+    control.setAttribute("data-target", `#${id}`);
+    control.setAttribute("aria-controls", id);
+    control.textContent = "Read more";
+
+    control.addEventListener("click", function () {
+      let isExpanded = control.getAttribute("aria-expanded") === "true";
+      control.textContent = isExpanded ? "Read more" : "Close";
+      control.setAttribute("aria-expanded", !isExpanded);
+      element.classList.toggle("show");
```
dnoneill, line 23: removing this line will fix this JS.

app/javascript/leaflet/controls/geosearch.js
```diff
@@ -0,0 +1,137 @@
...
+      ) {
+        return true;
+      } else {
+        return null;
+      }
+    };
+
+    const staticSearcher = (e) => {
+      setButtons();
+      if (staticButtonExists() && e.target.id == "gbl-static-button") {
```
dnoneill, line 47: Aren't lines 47-51 just doing what setButtons() is doing? I am not quite sure why this exists.

dnoneill: Also this js is pretty broken. When you zoom it will refresh the page.

app/javascript/leaflet/maps/bookmarks_map.js
```diff
@@ -0,0 +1,68 @@
...
+    let bbox;
+
+    if (typeof data.mapGeom === "string") {
+      bbox = geoJSONToBounds(JSON.parse(data.mapGeom));
+    } else {
+      document
+        .querySelectorAll(".document [data-geom]")
+        .forEach((docElement) => {
+          try {
+            const geomData = JSON.parse(docElement.dataset.geom);
```
dnoneill, line 22: Old code gets this from $(this).data()

app/javascript/leaflet/maps/homepage_map.js
```diff
@@ -0,0 +1,57 @@
+import GeoBlacklightViewerMap from "../viewers/map";
```
dnoneill, line 1: There is something kind of screwy here. Might be why the map isn't working but I am not sure.

app/javascript/leaflet/maps/search_results_map.js
```diff
@@ -0,0 +1,150 @@
...
+    // Create the span element for the glyphicon
+    const span = document.createElement("span");
+    span.className = "glyphicon glyphicon-repeat";
+
+    // Append the span to the anchor element
+    staticButtonNode.appendChild(span);
+
+    // DynamicButton
+    const dynamicButtonNode = document.createElement("label");
+    dynamicButtonNode.setAttribute("id", "gbl-dynamic-button");
```
dnoneill, line 34: I am not sure we should be redoing the dynamic / static button everywhere unless the button is different from the default.

app/javascript/leaflet/maps/search_results_map.js
```diff
@@ -0,0 +1,150 @@
...
+      baseUrl: "",
+      searcher: function () {
+        window.location.href = this.getSearchUrl();
+      },
+      delay: 800,
+      staticButton: staticButtonNode,
+      dynamicButton: dynamicButtonNode,
+    };
+
+    if (typeof data.mapGeom === "string") {
```
dnoneill, line 57: Lines 57-77 are the same as bookmarks. I wonder if we should create a function for them.

app/javascript/leaflet/maps/search_results_map.js
```diff
@@ -0,0 +1,150 @@
...
+            bbox =
+              typeof bbox === "undefined"
+                ? currentBounds
+                : bbox.extend(currentBounds);
+          } catch (e) {
+            bbox = bboxToBounds("-180 -90 180 90");
+          }
+        });
+    }
+
```
dnoneill, line 78: This is missing a history statechange https://github.com/geoblacklight/geoblacklight/blob/24aba2e82f4b386e0301d8beebf969727bcf7fb0/app/assets/javascripts/geoblacklight/modules/results.js#L2-L9 https://github.com/geoblacklight/geoblacklight/blob/24aba2e82f4b386e0301d8beebf969727bcf7fb0/app/assets/javascripts/geoblacklight/modules/results.js#L37-L45 not 100% sure what it is doing and maybe it got removed because all browsers support his now?

app/javascript/leaflet/maps/search_results_map.js
```diff
@@ -0,0 +1,150 @@
...
+    // instantiate new viewer
+    this.viewer = new GeoBlacklightViewerMap(element, { bbox });
+
+    // add geosearch control to viewer map
+    this.viewer.map.addControl(new GeoSearchControl(initialOptions));
+
+    // set hover listeners on map
+    document.getElementById("content").addEventListener(
```
dnoneill, line 86: Bookmarks has this same code in 86-107 as well. We might want to create a class that this extends.

app/javascript/leaflet/utils.js
```diff
@@ -0,0 +1,46 @@
...
+  sw = sw.wrap();
+  ne = ne.wrap();
+  return [
+    Util.formatNum(sw.lng, 6),
+    Util.formatNum(sw.lat, 6),
+    Util.formatNum(ne.lng, 6),
+    Util.formatNum(ne.lat, 6),
+  ];
+};
+
```
dnoneill, line 32: utils used to have a function called linkify. Not sure if linkify-html replaces it.

app/javascript/leaflet/viewers/esri/dynamic_map_layer.js
```diff
@@ -0,0 +1,110 @@
...
+    const pathArray = this.data.url.split("/");
+    const lastSegment = pathArray[pathArray.length - 1];
+
+    // If the last segment is an integer, slice and save as dynamicLayerId
+    if (Number(lastSegment) === parseInt(lastSegment, 10)) {
+      this.dynamicLayerId = lastSegment;
+      this.data.url = this.data.url.slice(0, -(lastSegment.length + 1));
+    }
+
+    get(this.data.url, {}, (error, response) => {
```
dnoneill, line 21: probably be good to have some consistency in esri.js we have import { get as esriGet } from "esri-leaflet";

app/javascript/leaflet/viewers/esri/dynamic_map_layer.js
```diff
@@ -0,0 +1,110 @@
...
+        if (error || response.results < 1) {
+          this.appendErrorMessage();
+        } else {
+          this.populateAttributeTable(featureCollection.features[0]);
+        }
+      });
+    });
+  }
+
+  appendLoadingMessage() {
```
dnoneill, line 82: Everything from here to the end is the in the esri.js file already. I am pretty sure we should be able to get rid of it.

app/javascript/leaflet/viewers/esri/feature_layer.js
```diff
@@ -0,0 +1,106 @@
...
+          if (error || !featureCollection.features.length) {
+            this.appendErrorMessage();
+          } else {
+            this.populateAttributeTable(featureCollection.features[0]);
+          }
+        });
+    });
+  }
+
+  appendLoadingMessage() {
```
dnoneill, line 74: Everything from here to the end is the in the esri.js file already. I am pretty sure we should be able to get rid of it.

app/javascript/leaflet/viewers/esri/feature_layer.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  getFeatureStyle() {
+    // Lookup style hash based on layer geometry type and return function
+    return (feature) => this.defaultStyles[this.layerInfo.geometryType];
+  }
+
+  setupInitialOpacity(featureLayer) {
+    featureLayer.on("load", () => {
+      featureLayer.setOpacity(this.options.opacity || 1); // Assuming `this.options.opacity` is defined
```
dnoneill, line 48: This is working. I am not sure why, but it looks like  this.esriFeatureLayer.setOpacity isn't setting a function the way we expect.

app/javascript/leaflet/viewers/esri/feature_layer.js
```diff
@@ -0,0 +1,106 @@
+import { featureLayer } from "esri-leaflet";
```
dnoneill, line 1: This is related to https://github.com/geoblacklight/geoblacklight/pull/1392/files#diff-fee17ec305e10d4416e6dfa4e2b2ad6223d976b376ee0bf6c1d0fbf4c447a1e2R12 but for some reason the http://localhost:3000/catalog/f406332e63eb4478a9560ad86ae90327_18 is not lading markers correctly and it thinks they are http://localhost:3000/catalog/../marker-icon.png which isn't anything.

app/javascript/leaflet/viewers/esri/tiled_map_layer.js
```diff
@@ -0,0 +1,73 @@
...
+          if (error) {
+            this.appendErrorMessage();
+          } else {
+            this.populateAttributeTable(featureCollection.features[0]);
+          }
+        });
+    });
+  }
+
+  appendLoadingMessage() {
```
dnoneill, line 51: Same as other comments

app/javascript/leaflet/viewers/esri/tiled_map_layer.js
```diff
@@ -0,0 +1,73 @@
+import { tiledMapLayer, identifyFeatures } from "esri-leaflet";
```
dnoneill, line 1: http://localhost:3000/catalog/nyu-test-soil-survey-map is now a feature layer, not a tiled map which means there is no way to test this, also I feel like if the info is in the esri like this https://dnrmaps.wi.gov/arcgis2/rest/services/TS_AGOL_STAGING_SERVICES/EN_AGOL_STAGING_SurfaceWater_WTM/MapServer/2 we should probably update it.

app/javascript/leaflet/viewers/esri.js
```diff
@@ -0,0 +1,85 @@
...
+      <tr><td colspan="2">Could not find that feature</td></tr></tbody>`;
+  }
+
+  populateAttributeTable(feature) {
+    let html = '<tbody class="attribute-table-body">';
+
+    // Step through properties and append to table
+    Object.keys(feature.properties).forEach((property) => {
+      html += `<tr><td>${property}</td>
+               <td>${linkifyHtml(feature.properties[property])}</td></tr>`;
```
dnoneill, line 76: linkifyhtml causes an error in http://localhost:3000/catalog/02236876-9c21-42f6-9870-d2562da8e44f, I am guessing because if it expects html? I am guessing this relates back to that missing linkify function. We might want to use that instead.

app/javascript/leaflet/viewers/iiif.js
```diff
@@ -0,0 +1,36 @@
...
+    this.adjustLayout();
+
+    this.map = map(this.element, {
+      center: [0, 0],
+      crs: CRS.Simple,
+      zoom: 0,
+    });
+
+    this.iiifLayer = tileLayer.iiif(this.data.url).addTo(this.map);
+    this.addControls();
```
dnoneill, line 21: Just leaving a note that 20 and 21 are switched in the original code but 99.999% sure it doesn't matter.

app/javascript/leaflet/viewers/index_map.js
```diff
@@ -0,0 +1,176 @@
...
+
+    if (this.available) {
+      this.addPreviewLayer();
+      this.addControls();
+    } else {
+      this.addBoundsOverlay(this.options.bbox);
+    }
+  }
+
+  indexMapInfoTemplate = (data) => `
```
dnoneill, line 17: I think this can be split into a function. I think you can create a function call docMetadata that takes a field

def docMetadata(field, fieldlabel, link=false) {
if (!field) { return }
var html = `<dt class="col-sm-3">${fieldlabel}:</dt>`
if (link){
html += `<dt class="col-sm-3"><a href="{{field}}">{{field}}</a></dt>`
} else {
html += `<dd class="col-sm-9">{{field}}</dd>`
}

app/javascript/leaflet/viewers/index_map.js
```diff
@@ -0,0 +1,176 @@
...
+
+      const indexMapDownloadHtml =
+        this.indexMapDownloadTemplate(updatedProperties);
+      const downloadElement = document.querySelector(".js-download-list");
+
+      if (Boolean(downloadElement)) {
+        downloadElement.innerHTML += indexMapDownloadHtml; // Appends new download template
+        downloadElement
+          .querySelectorAll(".js-index-map-feature")
+          .forEach((elem) => elem.remove());
```
dnoneill, line 172: old code did this before appending indexMapDownloadHtml to js-index-map-feature.

app/javascript/leaflet/viewers/map.js
```diff
@@ -0,0 +1,69 @@
...
+      ...options,
+    };
+
+    this.overlay = layerGroup();
+    this.onLoad();
+  }
+
+  onLoad() {
+    if (this.data.mapGeom) {
+      this.options.bbox = geoJSONToBounds(JSON.parse(this.data.mapGeom));
```
dnoneill, line 26: everything going into geoJSONToBounds seems to need JSON.parse. I am going to suggest we have geoJSONToBounds check the type of object going into it and do a JSON.parse if it is a string.

app/javascript/leaflet/viewers/map.js
```diff
@@ -0,0 +1,69 @@
...
+    if (this.map.getBounds().isValid()) {
+      this.element.setAttribute(
+        "data-js-map-render-bbox",
+        this.map.getBounds().toBBoxString()
+      );
+    }
+
+    this.map.addLayer(this.selectBasemap());
+    this.map.addLayer(this.overlay);
+  }
```
dnoneill, line 41: if (this.data.map !== 'index') {
      this.addBoundsOverlay(this.options.bbox);
    }

app/javascript/leaflet/viewers/map.js
```diff
@@ -0,0 +1,69 @@
...
+
+  addGeoJsonOverlay(geojson) {
+    const layer = geoJSON();
+    layer.addData(geojson);
+    this.overlay.addLayer(layer);
+  }
+
+  selectBasemap() {
+    return this.data.basemap ? basemaps[this.data.basemap] : basemaps.positron;
```
dnoneill, line 67: this.data isn't anything I don't think. At least I don't see it defined anywhere

app/javascript/leaflet/viewers/oembed.js
```diff
@@ -0,0 +1,21 @@
+import LeafletViewerBase from "./base.js";
+
+export default class LeafletViewerOembed extends LeafletViewerBase {
+  constructor() {
+    super();
```
dnoneill, line 5: Need to add constructor(element) + super(element). Currently not working. http://localhost:3000/catalog/stanford-dc482zx1528

app/javascript/leaflet/viewers/tilejson.js
```diff
@@ -0,0 +1,31 @@
...
+      })
+      .catch((error) => console.debug(error));
+  }
+
+  getBounds(doc) {
+    const bounds = doc.bounds;
+    if (bounds) {
+      const corner1 = latLng(bounds[1], bounds[0]),
+        corner2 = latLng(bounds[3], bounds[2]);
+      return latLngBounds(corner1, corner2);
```
dnoneill, line 28: seems to be working but looks like it isn't? http://localhost:3000/catalog/princeton-fk4544658v-tilejson shows all black tiles, but https://map-tiles-staging.princeton.edu/2a91d82c541c426cb787cc62afe8f248/mosaicjson/tiles/WebMercatorQuad/5/18/12@1x returns those black tiles so I think this is working.

app/javascript/leaflet/viewers/wms.js
```diff
@@ -0,0 +1,81 @@
+import { tileLayer } from "leaflet";
+import LeafletViewerMap from "./map.js";
+
+export default class LeafletViewerWms extends LeafletViewerMap {
+  onLoad() {
+    super.onLoad();
```
dnoneill, line 6: This function has additional code not in the original but I am pretty sure that is fine.
```
if (this.map.getBounds().isValid()) {
      this.element.setAttribute(
        "data-js-map-render-bbox",
        this.map.getBounds().toBBoxString()
      );
    }
 ```

File: app/javascript/leaflet/viewers/wms.js [hunk 33]
```diff
@@ -0,0 +1,81 @@
...
+        if (data.hasOwnProperty("error") || data.values.length === 0) {
+          document.querySelector(".attribute-table-body").innerHTML =
+            '<tr><td colspan="2">Could not find that feature</td></tr>';
+          return;
+        }
+
+        const html = document.createElement("tbody");
+        html.className = "attribute-table-body";
+        data.values.forEach((val) => {
+          html.innerHTML += `<tr><td>${val[0]}</td><td>${val[1]}</td></tr>`; // Note: You might need to implement linkify functionality
```
dnoneill, line 73: Old code had GeoBlacklight.Util.linkify(val[1])

app/javascript/leaflet/viewers/wmts.js
```diff
@@ -0,0 +1,81 @@
...
+      ? styleElement.getElementsByTagName("ows:Identifier")[0].textContent
+      : undefined;
+  }
+
+  getTileMatrixSet(layer) {
+    const tileMatrixElement = layer.getElementsByTagName("TileMatrixSet")[0];
+    return tileMatrixElement ? tileMatrixElement.textContent : undefined;
+  }
+
+  getBounds(layer) {
```
dnoneill, line 70: Old code has comment 

```
    // Get the layer's lower corner and upper corner values,
    // if they exist, and convert to a Leaflet latlngBounds object
```

app/javascript/leaflet/viewers/esri/feature_layer.js
dnoneill: I see now, marker-icon was in vendor/assets/images/marker-icon.png. Don't know why my github search didn't surface that. We are going to need to figure out where to put them.

app/javascript/leaflet/viewers/esri/dynamic_map_layer.js
thatbudakguy: For some reason the implementations are slightly different (one creates a new DOM node and one doesn't??) but in any case you're right – I'll just delete this one

app/javascript/leaflet/viewers/esri/dynamic_map_layer.js
thatbudakguy: I changed the other one; I think just `get` is fine

dnoneill:
The logo in the navbar isn't loading. It is due to the fact that we need to be able to point to something in the asset pipeline, without sprockets we need to use `url(asset-path('logo.svg'))` however I am not quite sure how to get vite to compile the right url. https://github.com/geoblacklight/geoblacklight/blob/dfbfe52984a4feda6749dd56e7b8c8af69c4cb67/app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss#L5

peetucket:
Reminder: add `npm run build` or similar and what to do when changing JS in development

eliotjordan (changes requested):
Some initial observations:

1. PMTiles record: The viewer div is way too large and the map isn't centered.
Screenshot 2024-07-26 at 9 58 01 AM
2. Clover Viewer: Images aren't centered on initial load.
Screenshot 2024-07-26 at 9 58 16 AM
Screenshot 2024-07-26 at 9 59 52 AM
3. COG Layer: Image didn't load at all, and think the container div is larger than it should be.
Screenshot 2024-07-26 at 9 58 49 AM

In addition, when I try to generate an app with this branch I get several errors. You can replicate this by replacing `gem "geoblacklight"` to `gem "geoblacklight", github: "geoblacklight/geoblacklight", branch: "modernize-js"` in the template.rb file and the running: `rails new test-gbl-js -m /path./to/template.rb`
1. Sprockets::DoubleLinkError in Catalog#index
```
Multiple files with the same output path cannot be linked ("application.js")
In "/Users/eliotj/Documents/Temp/test-gbl-js/app/assets/config/manifest.js" these files were linked:
  - /Users/eliotj/Documents/Temp/test-gbl-js/app/assets/javascripts/application.js
  - /Users/eliotj/Documents/Temp/test-gbl-js/app/javascript/application.js
```
2. I can rename `app/javascript/application.js` to something like `app/javascript/application2.js` and then I get ViteRuby::MissingEntrypointError in Catalog#index.
```
Vite Ruby can't find entrypoints/application.scss in the manifests.

Possible causes:
  - The last build failed. Try running `bin/vite build --clear --mode=development` manually and check for errors.
  ...
  
  Error: Can't find stylesheet to import.
     ╷
  11 │ @import '../../../node_modules/@geoblacklight/frontend/dist/stylesheets/geoblacklight/geoblacklight.scss';
     │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

There maybe other issues, but these are what I observed initially.

template.rb
```diff
@@ -2,15 +2,6 @@
...
 gem "geoblacklight", "~> 4.0"
-gem "sprockets", "< 4.0" # Use sprockets less than 4.0, let webpacker users set this up themselves
-
-# Hack for https://github.com/rails/rails/issues/35153
-# Adapted from https://github.com/projectblacklight/blacklight/pull/2065
-gemfile = File.expand_path("Gemfile")
-File.write(gemfile, File.open(gemfile) do |f|
-  text = f.read
-  text.gsub(/^gem 'sqlite3'$/, 'gem "sqlite3", "~> 1.3.6"')
-end)
```
eliotjordan, line 13: issue: we should leave this sqlite logic here. My testing shows this is still an issue.

eliotjordan: Hmmm. now I'm note sure. Resolving for now.

thatbudakguy:
Here's how to see this working using the `template.rb` build.

1. `rm -rf .internal_test_app`, just in case
2. `rails new .internal_test_app -m ./template.rb -a propshaft`; note that sprockets won't work so you need to use propshaft
3. `cd .internal_test_app && yarn add @geoblacklight/frontend@latest`; the app will build with the frontend assets matching the current gem version which is still v4.4; this will give you the latest frontend assets
4. run the server with `bin/rails s` (along with solr) and verify that it loads

eliotjordan (approved):
Excellent work. 💯

marlo-longley:
Yayyy!! 🥳 Great work everyone.

PR #1400, opened 2024-07-16
move header to components

### LINKED ISSUE #1396: ViewComponent – Header
We currently use the `shared/_header_navbar` partial to do this. As of Blacklight 7, [that partial just renders a view component](https://github.com/projectblacklight/blacklight/blob/release-7.x/app/views/shared/_header_navbar.html.erb#L1), so we can:

- Remove our override of the partial
- Create a new view component that includes the HTML currently in the partial
- Make sure that `blacklight_config.header_component` is set to our new view component

dnoneill:
closes #1396 

**Quite note on the homepage the search bar isn't exactly the same. I had to change the col-md-6 to col-md-10 because 6 didn't work well with the flex wrapper around the component search bar.**

Before | After - at 1344 width
Screenshot 2024-07-16 at 5 23 35 PM Screenshot 2024-07-16 at 5 23 35 PM

Before | After
Screenshot 2024-07-16 at 12 55 50 PM Screenshot 2024-07-16 at 5 10 02 PM

Before | After

Screenshot 2024-07-16 at 12 55 58 PM Screenshot 2024-07-16 at 5 10 10 PM


app/views/shared/_header_navbar.html.erb
```diff
@@ -1,46 +0,0 @@
-<nav class="navbar navbar-expand-md navbar-dark bg-dark topbar" role="navigation">
-  <div class="<%= container_classes %>">
-    <%= link_to application_name, root_path, class: 'mb-0 navbar-logo' %>
-    <button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#user-util-collapse" aria-controls="user-util-collapse" aria-expanded="false" aria-label="Toggle navigation">
-      <span class="navbar-toggler-icon"></span>
-    </button>
-
-    <div class="collapse navbar-collapse justify-content-md-end" id="user-util-collapse">
-      <%= render 'shared/user_util_links' %>
```
dnoneill, line 9: Pretty sure we still have this with https://github.com/projectblacklight/blacklight/blob/deb5c53b20552b4cba148b0002844e14439acf59/app/components/blacklight/top_navbar_component.html.erb

app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
```diff
@@ -45,3 +45,7 @@ $logo-image: image_url('blacklight/logo.svg') !default;
   padding-top:0.5rem;
 }
 // End - Addresses GBL issue #639
+
+.navbar-search.bg-light {
+  background-color: inherit!important;
+}
```
thatbudakguy, line 51: why the `!important`? does something else end up overriding this later?

app/components/geoblacklight/header_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class HeaderComponent < Blacklight::HeaderComponent
+    def initialize(blacklight_config:)
+      @blacklight_config = blacklight_config
+    end
```
thatbudakguy, line 7: we can leave this out entirely because it's identical in the parent, no?

app/components/geoblacklight/header_component.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= top_bar %>
+<% if controller_name == 'catalog' && (helpers.has_search_parameters? || params[:action] == 'show') %>
```
jcoyne, line 2: It would be nice to put this into a method like `show_search_bar?`

app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
dnoneill: Yeah the default in blacklight is a whitish color.

app/components/geoblacklight/header_component.html.erb
dnoneill: agreed.

app/components/geoblacklight/header_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class HeaderComponent < Blacklight::HeaderComponent
+    def not_landing_page?
+      controller_name == "catalog" && (helpers.has_search_parameters? || params[:action] == "show")
```
jcoyne, line 6: I know this conditional already existed, so no need to change in this PR, but in general testing on the `controller_name == 'catalog'` is suboptimal because it's okay to give the controller powered by blacklight a different name.  It should not be required to be called `CatalogController`

eliotjordan (approved):
Looks good.

Issue #1406, opened 2024-07-17
Updated indication for preparing download


peetucket:
Proxy ticket for https://github.com/sul-dlss/earthworks/issues/1092

Basically, provide a visible indicator to users that a file is being processed for download since this can take more than a few seconds.

peetucket:
Fixed by #1409 (pending the merge of modernize-js branch into main, since that work was done on the modernize-js branch)

PR #1407, opened 2024-07-17
improved download indicator

peetucket:
Attempt at fixing #1406.  May benefit from further UI refinements (like what is shown in https://github.com/sul-dlss/earthworks/issues/1092#issuecomment-2234348441)

Here is what it looks like with this PR:

download

### DIFF
--- a/app/assets/javascripts/geoblacklight/downloaders/downloader.js
+++ b/app/assets/javascripts/geoblacklight/downloaders/downloader.js
@@ -10,7 +10,7 @@
       L.Util.setOptions(this, options);
       this.$el = $(el);
       this.options.spinner.hide();
-      $('.exports .card-header').append(this.options.spinner);
+      this.$el.before(this.options.spinner);
       this.configureHandler();
     },
 
@@ -26,6 +26,9 @@
       this.downloading = true;
       var url = this.$el.data('downloadPath');
       this.options.spinner.show();
+      this.$el.removeAttr('href');
+      this.$el.prop('disabled', true);
+      this.$el.prop('innerHTML', 'Preparing download...');
       $.getJSON(url)
         .done(L.Util.bind(this.complete, this))
         .fail(L.Util.bind(this.error, this));
@@ -34,12 +37,14 @@
     complete: function(data) {
       this.downloading = false;
       this.$el.prop('disabled', false);
+      this.$el.prop('outerHTML', data[0][1]);
       this.renderMessage(data);
       this.options.spinner.hide();
     },
 
     error: function(data) {
       this.downloading = false;
+      this.$el.prop('innerHTML', 'Download failed');
       this.$el.prop('disabled', false);
       this.options.spinner.hide();
     },


app/assets/javascripts/geoblacklight/downloaders/downloader.js
```diff
@@ -10,7 +10,7 @@
       L.Util.setOptions(this, options);
       this.$el = $(el);
       this.options.spinner.hide();
-      $('.exports .card-header').append(this.options.spinner);
+      this.$el.before(this.options.spinner);
```
peetucket, line 13: the spinner was never appearing for me, I think it is because I couldn't find a DOM element with the `exports` class in it .... this instead places it near the download link

app/assets/javascripts/geoblacklight/downloaders/downloader.js
```diff
@@ -26,6 +26,9 @@
       this.downloading = true;
       var url = this.$el.data('downloadPath');
       this.options.spinner.show();
+      this.$el.removeAttr('href');
+      this.$el.prop('disabled', true);
+      this.$el.prop('innerHTML', 'Preparing download...');
```
peetucket, line 31: this:

1. removes the href attribute, which removes the link styling
2. disables the link (probably unnecessary, but it looks like existing JS code was re-enabling already when downloading was done)
3. changes the text of the link to indicate the download is processing

app/assets/javascripts/geoblacklight/downloaders/downloader.js
```diff
@@ -34,12 +37,14 @@
     complete: function(data) {
       this.downloading = false;
       this.$el.prop('disabled', false);
+      this.$el.prop('outerHTML', data[0][1]);
```
peetucket, line 40: this replaces the `Preparing download` message with the actual link to the download itself when it is ready ... it is in the response that is also used to generate the flash alert message 

response from server looks something like this, coming from here: https://github.com/geoblacklight/geoblacklight/blob/main/app/controllers/download_controller.rb#L27-L30

```
[
    [
        "success",
        "\u003ca data-download=\"trigger\" data-download-id=\"stanford-cz128vq0535\" data-download-type=\"generated-geojson\" href=\"/download/file/stanford-cz128vq0535-geojson.json\"\u003eYour file stanford-cz128vq0535-geojson.json is ready for download\u003c/a\u003e"
    ]
]
```

peetucket: Could also just make this a simple message like, and the user would have to get the actual link from the alert (as they do now).  OR, we could create a new link with any text we want (e.g. link the text "Download ready" to the file itself) though doing this would require adding some extra data to the response from the server to include just the link itself in addition to the full flash message.

```
  this.$el.prop('innerHTML', 'Download ready');
```

peetucket:
Have an alternative change that makes the link text for downloading the file when ready to "Download ready" instead, which looks a bit cleaner (though it required changing the json response from the server and a couple other adjustments):

Screenshot 2024-07-17 at 3 39 56 PM

app/assets/javascripts/geoblacklight/downloaders/downloader.js
dnoneill: So disabled doesn't work on href, this does work: https://github.com/geoblacklight/geoblacklight/pull/1404/commits/6e59a236526d7631e092eb7470fb999ca56b112e

dnoneill:
I think making it clean download ready for what type is important.

app/assets/javascripts/geoblacklight/downloaders/downloader.js
peetucket: thanks - i'll make this change in my PR (and also remove the need to take away the href attr, which i don't think will be needed anymore

thatbudakguy: we do want to take away the href, because that's the best way to indicate to screen reader users that it's not really a link. doing it with CSS won't tell them that, and adding `disabled` doesn't do anything.

dnoneill: I would just remove all the disabled stuff from code. It is not necessary anymore when the spinner replaces the link.

peetucket: I didn't have the spinner replacing the link in this PR, it was going next to it.  But now I also realize that this PR is built on top of main, and that the JS changed.... so I think I need to go to the `modernize-js` branch and work off of that instead?

thatbudakguy: It's kind of a tricky situation because that branch is a work in progress. But if you don't mind rebasing to that branch, it'd be super helpful

peetucket:
closed in favor of #1409

PR #1409, opened 2024-07-18
improved download indicator v2

peetucket:
Another attempt at fixing https://github.com/geoblacklight/geoblacklight/issues/1406. May benefit from further UI refinements (like what is shown in https://github.com/sul-dlss/earthworks/issues/1092#issuecomment-2234348441)

Possibly use Bootstrap toasts: https://getbootstrap.com/docs/4.6/components/toasts/

This PR takes the changes from #1407 and ports them into the modernized JS branch, while also moving the download javascript code to stimulus.

Changes:
1. Disables the "Export" link you clicked and instead shows a message indicating the download is preparing.
2. Shows a spinner to indicate activity is happening.
3. When download is ready, "Export" link is replaced with a link to download the file, labelled appropriately ... in addition, the download begins immediately (and the spinner is hidden).
4. The flash alert message with the link to the download is also shown at the top of the screen as before.
5. If the file is in the cache (has been generated recently), all of these steps will still happen, they will just happen very quickly, since we don't need to wait for the file to be generated.

How it now looks in this PR:

download


app/components/geoblacklight/download_links_component.rb
```diff
@@ -14,5 +14,78 @@ def initialize(document:, downloadable:)
...
+
+    # Format labels are customized for exports - look up the appropriate key.
+    def export_format_label(format)
+      t("geoblacklight.download.export_#{format.to_s.parameterize(separator: "_")}_link")
+    end
+
+    # Looks up properly formatted names for formats
+    def proper_case_format(format)
+      t("geoblacklight.formats.#{format.to_s.parameterize(separator: "_")}")
+    end
```
peetucket, line 89: these methods are all moved from `helpers/geoblacklight_helper.rb` since they aren't used anywhere except code that is now in the `DownloadLinksComponent`

app/components/geoblacklight/item_map_viewer_component.rb
```diff
@@ -25,7 +25,7 @@ def render?
     private
 
     def protocol
-      @document.viewer_protocol.camelize if @document.viewer_protocol
+      @document.viewer_protocol&.camelize
```
peetucket, line 28: to make linter happy

app/controllers/download_controller.rb
```diff
@@ -25,8 +25,8 @@ def show
     response = check_type
     validate response
     respond_to do |format|
-      format.json { render json: flash, response: response }
-      format.html { render json: flash, response: response }
+      format.json { render json: [flash, download_file_path(response)], response: response }
+      format.html { render json: [flash, download_file_path(response)], response: response }
```
peetucket, line 29: change the response from the server so that we have the link to download the file in the response...this allows us to construct a download link to replace the link that was used to generate the file (in addition to the showing the flash message)

app/helpers/geoblacklight_helper.rb
```diff
@@ -108,15 +44,6 @@ def formatted_name_reference(reference)
...
-  ##
-  # Wraps download text with proper_case_format
-  #
-  def download_text(format)
-    download_format = proper_case_format(format)
-    value = t("geoblacklight.download.download_link", download_format: download_format)
-    value.html_safe
-  end
-
   ##
```
peetucket, line 47: methods moved to the component

app/javascript/controllers/downloads_controller.js
```diff
@@ -0,0 +1,53 @@
...
+
+  renderMessage(message) {
+    Object.entries(message).forEach(([idx, msg]) => {
+      const flash = document.createElement("div");
+      flash.className = `alert alert-${msg[0]}`;
+      flash.innerHTML = `<button type="button" class="close" data-dismiss="alert" aria-hidden="true">&times;</button>${msg[1]}`;
+      document.querySelector("div.flash_messages").appendChild(flash);
+    });
+  }
+}
```
peetucket, line 53: new stimulus controller is mostly the same as the `downloader.js` file below, but adjusted to use stimulus targets and syntax

app/javascript/geoblacklight/downloaders/hgl_downloader.js
```diff
@@ -38,4 +37,21 @@ export default class GeoBlacklightHglDownloader extends GeoBlacklightDownloader
...
+    this.spinnerTarget.style.display = "none";
+  }
+
+  error(data) {
+    console.error("Download error:", data);
+    this.downloading = false;
+    this.element.classList.remove('disabled');
+    this.element.innerHTML = 'Download failed';
+    this.spinnerTarget.style.display = "none";
+  }
```
peetucket, line 56: the super class is now gone, but we still need this methods here

app/javascript/geoblacklight/downloads.js
```diff
@@ -1,11 +1,6 @@
-import GeoBlacklightDownloader from "./downloaders/downloader";
 import GeoBlacklightHglDownloader from "./downloaders/hgl_downloader";
 
 export default function initializeDownloads() {
-  document.querySelectorAll("[data-download-path]").forEach((element) => {
-    new GeoBlacklightDownloader(element);
-  });
-
```
peetucket, line 8: done in stimulus now

dbranchini:
I have two ideas for the UI for this. I believe the first one will be easier to implement:

**Idea 1** 

https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=911-24473&viewport=-7294%2C1192%2C0.56&t=kYki5n8Z1jZKPmeu-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=911%3A24473&show-proto-sidebar=1

Start at the Stanford public vector screen, click on the GeoJSON link to see it. 

**Idea 2 - a toast**

https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=971-31966&viewport=-7294%2C1192%2C0.56&t=kYki5n8Z1jZKPmeu-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=971%3A31966&show-proto-sidebar=1

Start at the Alt download ready screen, click on the GeoJSON link to see it. 

Both of these ideas require the user to get the item from their browser downloads once they see the ready.

peetucket:
Both of those options look good, I don't think either would be too hard to implement.  I think we wanted to hold off on using a toast until we upgrade to bootstrap 5, so we don't have to change around how it works.

One question is if we want to the download to happen automatically when it's ready in addition to providing the link in the "Download ready" message.  This PR does that (i.e. if effectively just automatically clicks the download link for you when the file is ready, which starts the download to your browser immediately).

peetucket:
Thinking about this in the the context of this morning's standup discussion on re-usable components: do we have ideas for how we'd implement something like the pulsing three dots for the "preparing download"?   Bootstrap has some spinner stuff but it's not quite what is shown there:   https://getbootstrap.com/docs/4.6/components/spinners/

If designs had pointers to implementations in a toolkit library (like what is started at https://sul-dlss.github.io/component-library/) this would greatly facilitate commonality between apps and also make development more rapid (so we don't each individually try and figure out how to implement a design mockup).

app/assets/stylesheets/geoblacklight/modules/downloads.scss
```diff
@@ -47,3 +47,9 @@
     }
   }
 }
+
+.disabled {
+  cursor: not-allowed;
+  opacity: 0.5;
+  text-decoration: none;
+}
```
thatbudakguy, line 55: could we make this `a.disabled` or something? seems like there's a chance it could conflict with other styles (bootstrap?) that have a more generic `disabled` class

peetucket:
> https://www.figma.com/proto/nkgg905s5FyXoyPZjVMco0/EarthWorks?page-id=0%3A1&node-id=911-24473&viewport=-7294%2C1192%2C0.56&t=kYki5n8Z1jZKPmeu-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=911%3A24473&show-proto-sidebar=1
> 
> Screenshot 2024-07-18 at 5 19 58 PM
> 

For option 1, when we show "Download ready", does it make sense to indicate which file is ready?  e.g. "Download ready (shapefile)"  The reason to do this is that some objects have multiple file types available, and if you clicked multiple, when each is ready they would otherwise all render the same "Download ready" text and you wouldn't know which was which.

peetucket:
Some additional styling added (uses spinner and checkmark when done).  See the GIF:

download

PR #1410, opened 2024-07-18
Migrate inspection javascript to controller pattern

marlo-longley:
@thatbudakguy this is mostly ready! Big thanks to @dnoneill!
The one issue we are having this error in rebasing to your new `addControls` code in the leaflet controller: `const protocolControls = this.config.CONTROLS || [];`

I believe it is broken in your branch that we rebased to. To keep the PR merge simple, I think it would be best if you fixed it in the `modernize-js` branch and then we rebased again here. Thoughts? Thanks!

In http://localhost:3000/catalog/cugir-007957:
```
index-CBL7Q53g.js:149452 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'CONTROLS')
    at extended.addControls (index-CBL7Q53g.js:149452:42)
    at extended.addPreviewOverlay (index-CBL7Q53g.js:149508:12)
```


app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -151,33 +155,47 @@ export default class LeafletViewerController extends Controller {
     if (this.boundsOverlay) this.overlay.removeLayer(this.boundsOverlay);
   }
 
+  addInspection() {
+    if (this.protocolValue == "Wms") return wmsInspection(this.map, this.urlValue, this.layerIdValue);
```
jcoyne, line 159: This looks like a good place to use `switch`

app/javascript/leaflet/utils.js
```diff
@@ -52,4 +52,27 @@ export const getTileJsonBounds = (doc) => {
...
+export const appendErrorMessage = () => {
+  document.querySelector(
+    ".attribute-table-body"
+  ).innerHTML = `<tbody class="attribute-table-body">
+    <tr><td colspan="2">Could not find that feature</td></tr></tbody>`;
+};
+
+// Looks for strings that could be URLs and wraps them in <a> tags
+export const linkify = (str) => {
+  const urlRegEx = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-]*)?\??(?:[\-\+=&;%@\.\w]*)#?(?:[\.\!\/\\\w]*))?)/g;
```
jcoyne, line 76: This seems like a very curious regex.  Do we want it to support basic auth?  It seems odd it requires 3-9 letters (no websocket 'ws')  for the protocol. It has `www` embedded in it but doesn't allow other subdomains?

jcoyne: Consider using:
```javascript
try {
  url = URL.new(str.toString()).toString()
  return `<a href='${url}'>${url}</a>`
 } catch(error) {
 return str.toString();
 }
 ```

app/javascript/controllers/leaflet_viewer_controller.js
thatbudakguy: I had that initially but it's actually more lines because each `case` has to be its own line. And, if anyone forgot to `return` or `break`, you get fun fall-through behavior. I kind of consider JS `switch` to be harmful, tbh

thatbudakguy: What this should ultimately be (imo) is just a lookup in a map that maps the protocol names to callables. Which can happen at some point soon

app/javascript/leaflet/utils.js
thatbudakguy: iirc this is old code and it's a vendorized version of a package that does this...maybe we should reevaluate or just use the package instead

jcoyne: Oh...

marlo-longley: Yes this was ported over from the original code. We had trouble loading a linkify node module tbh with the new setup tbh so having our own method was easier to access. Happy to reevaluate.

thatbudakguy:
I'm going to merge this and we can come back to the linkify thing, because there's probably a better solution there

PR #1414, opened 2024-07-22
fix search here button and geosearch map

dnoneill:


app/javascript/leaflet/controls/layer_opacity.js
```diff
@@ -9,6 +9,7 @@ export default class LayerOpacityControl extends Control {
     if (typeof layer.getLayers !== "undefined") {
       // Add first layer from layer group to options
       this.options.layer = layer.getLayers()[0];
+      this.options.layer = this.options.layer ? this.options.layer : layer;
```
thatbudakguy, line 12: this looks like a case for the [nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment)!

```suggestion
      this.options.layer ??= layer;
```

app/javascript/leaflet/controls/geosearch.js
```diff
@@ -79,11 +79,11 @@ export default class GeoSearchControl extends Control {
     if (this.options.dynamic) {
       this.dynamicButton.style.display = "block";
     }
-    this.applySearch();
+    this.applySearch(false);
   }
 
   // Update the URL with the current bounding box and go back to page 1
-  applySearch() {
+  applySearch(turbo=true) {
```
thatbudakguy, line 86: I realize it's less javascript-y, but could we do
```suggestion
  applySearch(opts={turbo: true}) {
```
so that the calls look like
```js
this.applySearch({ turbo: false })
```
which is a little more explicit about what's happening?
```js
this.applySearch(false)
```
is kinda cryptic

app/javascript/leaflet/controls/geosearch.js
```diff
@@ -99,7 +99,8 @@ export default class GeoSearchControl extends Control {
     console.log('navigating to', newUrl, 'with Turbo:', window.Turbo);
 
     // If Turbo Drive is active, do the new page navigation without a full page reload
-    if (window.Turbo) window.Turbo.visit(newUrl);
+    // Do full page reload for staticSearch button
```
thatbudakguy, line 102: imo this comment actually belongs at the call site where `turbo = false`

thatbudakguy (approved):

PR #1415, opened 2024-07-23
fix default map to have bounds overlay

dnoneill:
This is what the default map should look like if no protocol

Screenshot 2024-07-23 at 12 10 41 PM

### DIFF
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -71,7 +71,7 @@ export default class LeafletViewerController extends Controller {
 
     // If the data is available, add the preview and set up inspection
     // Otherwise just draw the bounds, if configured to do so
-    if (this.availableValue) {
+    if (this.availableValue && this.protocolValue) {
       await this.addPreviewOverlay();
       this.addInspection();
     } else if (this.drawInitialBoundsValue && this.bounds) {


app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -197,6 +197,7 @@ export default class LeafletViewerController extends Controller {
     if (protocol == "Wms") return wmsLayer(url, options);
     if (protocol == "Wmts") return await wmtsLayer(url, options);
     if (protocol == "Xyz") return tileLayer(url, options);
+    if (!protocol && this.bounds instanceof L.LatLngBounds) return L.polygon([this.bounds.getSouthWest(), this.bounds.getSouthEast(), this.bounds.getNorthEast(), this.bounds.getNorthWest()]);
```
thatbudakguy, line 200: isn't this exactly what `this.addBoundsOverlay()` does?

thatbudakguy: imo this should go in `loadMap()` or somewhere else and just call `this.addBoundsOverlay()`, since this function is about adding actual data to the map

dnoneill: I did not see that. Yes, so it looks like we aren't setting drawInitialBoundsValue, I will fix this.

PR #1420, opened 2024-07-23
fix download specs

peetucket:
(replaces #1417 which was borked due to a rebase)

Download specs are not working correctly in the modernize-js branch due to changes in the javascript from https://github.com/geoblacklight/geoblacklight/pull/1409

This fixes them.

### DIFF
--- a/app/controllers/download_controller.rb
+++ b/app/controllers/download_controller.rb
@@ -14,8 +14,8 @@ class DownloadController < ApplicationController
           download_type: "generated-#{params[:type]}"
         })
     respond_to do |format|
-      format.json { render json: flash, response: response }
-      format.html { render json: flash, response: response }
+      format.json { render json: [flash, ""], response: response, status: 500 }
+      format.html { render json: [flash, ""], response: response, status: 500 }
     end
   end
 
--- a/spec/features/download_layer_spec.rb
+++ b/spec/features/download_layer_spec.rb
@@ -11,14 +11,20 @@ feature "Download layer" do
     allow(Geoblacklight::KmzDownload).to receive(:new).and_return(kmz_download)
   end
 
+  # cleanup any downloaded files
+  after do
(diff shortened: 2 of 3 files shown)


spec/features/download_layer_spec.rb
```diff
@@ -11,14 +11,20 @@
...
+  after do
+    FileUtils.rm("tufts-cambridgegrid100-04-shapefile.zip", force: true)
+    FileUtils.rm("tufts-cambridgegrid100-04-kmz.kmz", force: true)
+  end
+
   scenario "clicking initial shapefile download button should trigger download", js: true do
-    expect(shapefile_download).to receive(:get).and_return("mit-f6rqs4ucovjk2-shapefile.zip")
-    visit solr_document_path("mit-f6rqs4ucovjk2")
+    expect(shapefile_download).to receive(:get).and_return("tufts-cambridgegrid100-04-shapefile.zip")
+    visit solr_document_path("tufts-cambridgegrid100-04")
```
peetucket, line 22: this tufts-cambridge fixture actually has files available to download in the fixtures, and since we auto-click to download, we should use fixtures where this works

app/controllers/download_controller.rb
```diff
@@ -14,8 +14,8 @@ class DownloadController < ApplicationController
           download_type: "generated-#{params[:type]}"
         })
     respond_to do |format|
-      format.json { render json: flash, response: response }
-      format.html { render json: flash, response: response }
+      format.json { render json: [flash, ""], response: response, status: 500 }
+      format.html { render json: [flash, ""], response: response, status: 500 }
```
peetucket, line 18: return a 500 response when there is an error, so our new downloads controller can correctly detect via status

also make the response from an exception match the response from a success (which changed to include the filename when available)

thatbudakguy (approved):

PR #1427, opened 2024-07-26
Fix viewer sizing for OpenLayers and IIIF viewers

thatbudakguy:
Setting a page-relative height with !important caused these viewers to be much larger than their container and have their content render off-center.

PMTiles:

Screenshot 2024-07-26 at 10 25 24

COG (still has a data problem?):

Screenshot 2024-07-26 at 10 25 19

IIIF (now centered):

Screenshot 2024-07-26 at 10 25 10

Notably, this doesn't fix the IIIF viewer (not the manifest one, just a single IIIF image). It seems to have some kind of height calculation problem. This might be a bug in Clover's single-image viewer, or just a configuration issue?

Screenshot 2024-07-26 at 10 23 43

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/item.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/item.scss
@@ -10,14 +10,10 @@
       border: 1px solid $gray-400;
       min-height: 27.5rem;
 
-      // Ensure oembed iframe is the same height as its container
-      iframe {
+      // Ensure embeds, openlayers viewer, and IIIF viewer fit their container
+      iframe, .ol-viewport, .clover-iiif-image-openseadragon {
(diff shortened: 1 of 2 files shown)


thatbudakguy:
@eliotjordan do you have any advice re: [Clover's `Image` component](https://samvera-labs.github.io/clover-iiif/docs/image)? that's what we're using to do the single IIIF image rendering, instead of leaflet-iiif.

thatbudakguy:
OK, I was able to fix it by telling React to insert a wrapper `<div>` with an inline style, as the Clover docs indicate for the `Image` component.

Screenshot 2024-07-26 at 11 06 44

eliotjordan (approved):

PR #1438, opened 2024-07-29
update map legend to work with setting colors

### LINKED ISSUE #1435: fix index map legend
### Expected behavior
Changing the color settings in https://github.com/geoblacklight/geoblacklight/blob/4af93ba5736bf53bbea143545f25d0955536b7ee/lib/generators/geoblacklight/templates/settings.yml#L234-L244 should update the map and index to match these colors. 

### Actual behavior

It does not, it keeps hard coded colors.
Screenshot 2024-07-29 at 12 38 07 PM

1. Update colors in  https://github.com/geoblacklight/geoblacklight/blob/4af93ba5736bf53bbea143545f25d0955536b7ee/lib/generators/geoblacklight/templates/settings.yml#L234-L244
2. start server

### Related work

Link to related tickets or prior related work here.

dnoneill:
closes #1435


spec/components/geoblacklight/index_map_legend_component_spec.rb
```diff
@@ -8,17 +8,17 @@
   end
 
   it "shows available map text" do
-    expect(rendered).to have_text("Green tile indicates")
+    expect(rendered).to have_text("#7FCDBB tile indicates")
```
marlo-longley, line 11: Can we add a color display_name to the settings? As a user, I don't know what the hex color corresponds to on the page. "Green" is more helpful.

marlo-longley (changes requested):

spec/components/geoblacklight/index_map_legend_component_spec.rb
dnoneill: I definitely had this thought but I was also wondering if the hex made more sense because this only shows up for screen readers and I would assume anyone using the screen reader might not necessarily be able to match green

marlo-longley (approved):

PR #1440, opened 2024-07-29
Add ability to run template.rb with a branch

eliotjordan:
Updates template.rb to allow installing development branches into a new rails app.

1. Uses branch ruby code.
2. Uses latest geoblacklight frontend package.

```
BRANCH=my_branch rails new app_name -m /path/to/template.rb
```


thatbudakguy (approved):
This worked and I got the latest version of the frontend package. 👍🏻

Issue #1292, opened 2023-06-06
Autosuggest - Migrate from Twitter Typeahead to GitHub AutoComplete


ewlarson:
### Descriptive summary

Blacklight has removed Twitter's [typeahead.js](https://twitter.github.io/typeahead.js/) library, which we've seen flagged with some accessibility issues in the B1G (missing field labels), and replaced the feature with [GitHub's AutoCompleteElement](https://github.com/github/auto-complete-element) library. We should follow suite in GBL.

### Related work

PR for library swap:
https://github.com/projectblacklight/blacklight/pull/2576

Commit removing typeahead:
https://github.com/projectblacklight/blacklight/commit/9d1c86f957678cebd10ac6d3b16c8843bc8f717c

Commit adding GH AutoComplete:
https://github.com/projectblacklight/blacklight/commit/882379cf36f2248e1c7b3a828b6e19cf1cdebf2a

Base layout implementation: 
https://github.com/projectblacklight/blacklight/blob/13a8122fc6495e52acabc33875b80b51613d8351/app/views/layouts/blacklight/base.html.erb#L22-L24

### Timing / Milestone

Maybe this work should wait until we tackle our official Blacklight v8 release?

PR #1441, opened 2024-07-30
Upgrade to Blacklight 8 and Bootstrap 5

### LINKED ISSUE #1373: Upgrade to Bootstrap 5
### Descriptive summary

This is a [framework recommendation](https://geoblacklight.org/docs/framework-recommendations/) for GBL v5, and also a requirement for upgrading to Blacklight 8.

### Related work

Work to move a lot of the views to View Components (see #1270) probably needs to be done first, otherwise there will be merge conflicts when the class names are changed in templates.

thatbudakguy:
This supersedes the earlier PR from which it was inspired, #1344.

Closes #1373 and #1292.


app/javascript/leaflet/inspection.js
```diff
@@ -6,7 +6,7 @@ export const wmsInspection = (map, url, layerId) => {
     const spinner = document.createElement("tbody");
     spinner.className = "attribute-table-body";
     spinner.innerHTML =
-      '<tr><td colspan="2"><div class="spinner-border" role="status"><span class="sr-only">Inspecting</span></div></td></tr>';
+      '<tr><td colspan="2"><div class="spinner-border" role="status"><span class="visaully-hidden">Inspecting</span></div></td></tr>';
```
marlo-longley, line 9: typo.

eliotjordan:
1. I've just now started to review this PR. First thing I notice after spinning up is how small and off center the search input is.
Screenshot 2024-07-31 at 6 27 26 PM

2. The map preview appears below the metadata section rather than above.
Screenshot 2024-07-31 at 6 31 20 PM

3. Also the default icon looks strange with the jet black background in the navbar.
Screenshot 2024-07-31 at 6 34 49 PM

thatbudakguy:
I'm working on 1 & 2.

thatbudakguy:
The header now matches BL8 and the viewer/attributes/etc. order is restored:

Screenshot 2024-08-01 at 14 13 11

marlo-longley:
@thatbudakguy looks great. Builds correctly. Thank you for all your work on this.
The one thing I am seeing is the background color of the homepage search bar -- assuming it was not intentional? I believe it should be white.
Screenshot 2024-08-01 at 17 47 49
Screenshot 2024-08-01 at 17 48 18

thatbudakguy:
That is blacklight 8's default.

Screenshot 2024-08-01 at 15 41 15

thatbudakguy:
The homepage jumbotron uses the same background color as the search bar. I thought it looked OK, so I left it 🤷🏻

app/components/geoblacklight/header_component.html.erb
```diff
@@ -1,18 +1,14 @@
-<%= top_bar %>
-<% if not_landing_page? %>
-<%= search_bar %>
-<% else %>
-<div class='geobl-homepage-masthead jumbotron'>
+<header>
```
jcoyne, line 1: Should we do this in Blacklight too? https://github.com/projectblacklight/blacklight/blob/main/app/components/blacklight/header_component.html.erb

jcoyne: In Argo, we didn't put the search_bar in the `<header>`, it went in top_bar instead. I wonder if we can find consistency on that. https://github.com/sul-dlss/argo/blob/main/app/components/top_navbar_component.html.erb#L1

thatbudakguy: I'm happy to adopt whatever Blacklight does 🤷🏻

eliotjordan (changes requested):
- Points 1,2 are much better. 
- I hadn't noticed the deafault logo issues before. It's not a huge deal since it's almost always replaced. The logo is an svg. I wonder if we can update the color. If not in this branch in another ticket.
- In the newest version of this branch, clicking on a feature in an index map does not update anything on the page and raises an error in the console. I haven't dug into it too far: 
```
Uncaught (in promise) 
TypeError: Cannot set properties of null (setting 'innerHTML')
```

thatbudakguy:
Good catch! the new `DocumentComponent` had left off a partial. I added it back and wrote some tests.

PR #1449, opened 2024-07-31
Build demo app in CI an push docker images

### LINKED ISSUE #1448: Add functionality to CI to build test apps and docker images
### Descriptive summary

It would be helpful for quick review of pull requests if we could perform the build step in CI and create docker images that can be pushed as artifacts. Additionally, if we use the template.rb file to generate these builds instead of `rake engine_cart:generate` it would allow us to test our template.

eliotjordan:
Build demo/test apps as artifacts and create docker images from them.

- On pull request and main branch push, build demo apps using the template.rb file.
- Generate a docker image from the test app and push into the GitHub package repository.
- Bundle a docker compose file with the downloadable build which runs solr and gbl app.
- Add comment to pull request with link for downloading the build.
- Prune old images monthly.
- Regenerate a new main branch image weekly.

Closes #1448

### DIFF
--- a/.github/workflows/build-demo-app.yml
+++ b/.github/workflows/build-demo-app.yml
@@ -1,0 +1,153 @@
+name: Build demo app
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "0 1 * * SUN"  # every sunday at 1AM
+  workflow_dispatch:
+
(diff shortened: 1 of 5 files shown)


ewlarson (approved):
🚀 Nice work! GBL/Solr stand up very easily on my end.

PR #1450, opened 2024-07-31
Add docker compose option for solr only and modify Rakefile

marlo-longley:
Downloading and building Solr and the rebuilding of internal test app so many times was a pain point for our dev team. This PR is meant to make development easier and is not related to production builds. It removes a redundant `engine_cart:generate` statement, adds an option to run Solr in docker, and makes it easier overall to run Solr and the application separately. I followed the pattern we used in Arclight development, another downstream-from-Blacklight rails engine, which has worked pretty well for us.

If approved, I have drafted updated documentation that I will PR in that repo.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -49,7 +49,7 @@ namespace :geoblacklight do
   end
 
   namespace :internal do
-    task seed: ["engine_cart:generate"] do
+    task :seed do
       within_test_app do
         system "bundle exec rake geoblacklight:index:seed"
         system "bundle exec rake geoblacklight:downloads:mkdir"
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.7"
```
eliotjordan, line 1: issue: The default name for this file is now `compose.yml` or `compose.yaml`. I would change the name to match.
https://docs.docker.com/compose/compose-application-model/#the-compose-file

eliotjordan (changes requested):
This is great! Seems compatible with the build and docker PR I posted earlier. Only thing is to change is the name of the compose file.

PR #1455, opened 2024-08-01
Rebuild base image manually or on a schedule

### LINKED ISSUE #1453: Rebuild base image once a week, not the main branch
To keep the size of the layers that docker compose pull needs to pull down, we should rebuild only the base image once a week. Not everytime the main branch is updated.

eliotjordan:
Closes #1453 

- Rebuild base image manually or on a schedule
- Fix delete image job and set to two week schedule


ewlarson (approved):
👍

PR #1458, opened 2024-08-02
Build and use javascript from gem instead of npm

### LINKED ISSUE #1457: When building demo app in CI, use javascript from gem
On a branch PR, in order to target correct versions of the gbl javascript, we should use the javascript from the gem instead of pulling from npm.

eliotjordan:
- Build and use javascript from gem instead of npm
- Renames a dirctory and a few files for clarity

Closes #1457


thatbudakguy (approved):
This seems like a good solution for `template.rb` builds 👍🏻

PR #1465, opened 2024-08-07
Center the search bar

marlo-longley:
According to the screenshots in the thread for PR #1441, the search bar should be centered, but correct me if I am wrong@!

### Before

Screenshot 2024-08-07 at 10 00 51
Screenshot 2024-08-07 at 10 09 16

### After

Screenshot 2024-08-14 at 13 53 11
Screenshot 2024-08-14 at 13 53 19

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
@@ -41,3 +41,8 @@ auto-complete {
 #main-container {
   padding-top: 1rem;
 }
+
+// center contents of the container that holds the search bar
+.navbar-search.navbar > .container {
+  justify-content: center;
+}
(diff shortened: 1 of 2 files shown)


dnoneill:
The search bar shouldn't be showing up on the homepage above the banner. It should be below the banner.

dnoneill:
Also the old main looked like this: https://github.com/geoblacklight/geoblacklight/pull/1400 so I think it was supposed to be off center?

marlo-longley:
@dnoneill Thanks for the reference. OK, based on that I see that the search results page should stay as is.
The homepage should 1. move the search bar beneath the "Explore and Discover" banner and 2. center the search bar.
If that sounds right I will update this PR.

eliotjordan:
@dnoneill Can we move the search bar below the Hero element text to match the existing GBL design?

marlo-longley:
@dnoneill @eliotjordan please see updated "After" screenshots in the top comment of this PR. Should be good!

eliotjordan (approved):

PR #1466, opened 2024-08-07
Add attribute inspection as sidebar on the map

### LINKED ISSUE #1576: Move inspection table to sidebar on the map
When you click on an attribute on the map, often you can miss them loading because they are located below the map and the map might be taking up the whole page. This moves the inspection to a sidebar so you can see everything without having to scroll.

dnoneill:
closes #1576. Move the full screen to the lower left hand corner because it looks weird with the sidebar 

### Changes
- add map data header that is collapsible
- add instructions to index maps
- moves full screen button below opacity bar

Screenshot 2025-07-30 at 2 50 46 PM
Screenshot 2025-07-30 at 2 50 43 PM
Screenshot 2025-07-30 at 2 50 38 PM

## With SIDEBAR set
Mobile view
Screenshot 2025-07-30 at 2 50 02 PM
Screenshot 2025-07-30 at 2 49 57 PM
Screenshot 2025-07-30 at 2 50 21 PM
Screenshot 2025-07-30 at 2 50 17 PM
Screenshot 2025-07-30 at 2 50 13 PM
Screenshot 2025-07-30 at 2 49 49 PM
Screenshot 2025-07-30 at 2 49 46 PM

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/item.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/item.scss
@@ -49,13 +49,30 @@
     -webkit-line-clamp: 8;
   }
 
-  #table-container{
-    overflow: auto;
-    max-height: 450px;
+  #table-container {
     margin-bottom: 1rem;
 
     table {
       margin-bottom: 0px;
+    }
+  }
+
+  .attribute-data {
+    --bs-accordion-active-bg: none;
+  }
+
+  .sidebar-container {
+    width: 25vw;
+    overflow: scroll;
+    overscroll-behavior: contain;
+
+    img {
+      width: 100%;
+    }
+
+    @media screen and (max-width: breakpoint-max(sm)) {
+      width: calc(100vw - 100px);
+      max-height: 20vh!important;
     }
   }
 }
--- a/app/components/geoblacklight/attribute_table_component.html.erb
+++ b/app/components/geoblacklight/attribute_table_component.html.erb
@@ -1,17 +1,15 @@
-<div class='row'>
-  <div id='table-container' class='col-md-12'>
-    <table id="attribute-table" class="table table-hover table-condensed table-striped table-bordered col-md-12">
-      <thead>
-        <tr>
-          <th><%= t('geoblacklight.show.attribute') %></th>
-          <th><%= t('geoblacklight.show.value') %></th>
-        </tr>
-      </thead>
-      <tbody class='attribute-table-body'>
-        <tr>
-          <td class='default-text' colspan='2'><em><%= t('geoblacklight.show.click_map') %></em></td>
-        </tr>
-      </tbody>
-    </table>
-  </div>
-</div>
+<%= render Geoblacklight::AccordionComponent.new(id: 'attribute-table', title: t('geoblacklight.show.table_name')) do  %>
+  <table id="attribute-table" class="table table-hover table-condensed table-striped mb-0 collapse show">
+    <thead>
+      <tr>
+        <th><%= t('geoblacklight.show.attribute') %></th>
+        <th><%= t('geoblacklight.show.value') %></th>
+      </tr>
+    </thead>
+    <tbody class='attribute-table-body'>
+      <tr>
+        <td class='default-text' colspan='2'><em><%= t('geoblacklight.show.click_map') %></em></td>
+      </tr>
+    </tbody>
+  </table>
+<% end %>
--- a/app/components/geoblacklight/index_map_inspect_component.html.erb
+++ b/app/components/geoblacklight/index_map_inspect_component.html.erb
@@ -1,4 +1,5 @@
-<div class='row'>
-  <div class='viewer-information col-sm-12'>
+<%= render Geoblacklight::AccordionComponent.new(id: 'index-map-attributes', title: t('geoblacklight.show.table_name')) do  %>
+  <div class='viewer-information px-3'>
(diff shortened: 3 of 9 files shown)


ewlarson:
I like this direction, if we can keep things manageable on smaller devices.

Also! Way back in the day we had talked about having a full feature table viewer. Something like this: https://developers.arcgis.com/javascript/latest/sample-code/widgets-featuretable-map/

Screenshot 2024-08-09 at 10 00 43 AM

srappel:
@ewlarson this would actually be amazing for index maps, because it would allow someone to, for example, browse by title and sort columns as needed.

ewlarson:
Yeah, we talked about that at today's stand up. We can't do it for WMS items, which could potentially have 1M rows (HT / @kgjenkins), but for index maps this could work.

app/assets/stylesheets/geoblacklight/modules/item.scss
```diff
@@ -45,13 +45,40 @@
...
     margin-bottom: 1rem;
 
     table {
       margin-bottom: 0px;
     }
   }
+
+  .sidebar-container {
+    width: 25vw;
+    overflow: scroll;
```
thatbudakguy, line 58:
```suggestion
    overflow: scroll;
    overscroll-behavior: contain;
```
I noticed that when scrolling in the sidebar, you could cause the underlying page to scroll if you scrolled past the end, which felt kind of strange. This seemed to fix it

app/components/geoblacklight/attribute_table_component.html.erb
```diff
@@ -1,6 +1,9 @@
-<div class='row'>
-  <div id='table-container' class='col-md-12'>
-    <table id="attribute-table" class="table table-hover table-condensed table-striped table-bordered col-md-12">
+<div id="<%= Settings.LEAFLET.SIDEBAR_ID || 'table-container' %>" data-sidebar="<%= Settings.LEAFLET.SIDEBAR_ID %>">
+  <div id='table-container' class="border bg-light<% if Settings.LEAFLET.SIDEBAR_ID %> sidebar-container<% end %>">
+    <button class="content-sidebar-header p-2 m-0 btn btn-link h6" type="button" aria-expanded="true" data-bs-toggle="collapse" data-bs-target="#attribute-table" aria-controls="attribute-table">
+      <%= t('geoblacklight.show.table_name') %>
+    </button>
```
thatbudakguy, line 5: This and the `IndexMapInspectComponent` now seem to have a lot of shared code around this collapsible portion. Could we extract that as a wrapper component, and use Bootstrap's [built-in accordion](https://getbootstrap.com/docs/5.3/components/accordion/) with a single item? I tried this locally and it seemed to work and look OK; you get the icon and the open/close behavior "for free".

app/components/geoblacklight/index_map_inspect_component.html.erb
```diff
@@ -1,4 +1,10 @@
-<div class='row'>
-  <div class='viewer-information col-sm-12'>
+<div id="<%= Settings.LEAFLET.SIDEBAR_ID || 'table-container' %>" data-sidebar="<%= Settings.LEAFLET.SIDEBAR_ID %>">
```
thatbudakguy, line 1: I think I'd rather have the simplicity/readability of a hardcoded/default sidebar ID than whatever this is allowing you to do. Judging from the javascript, if `SIDEBAR_ID` is set, we will find this element, look at its data attribute, and then use that to find another element...which is always itself? I'm not convinced the flexibility is necessary right now.

thatbudakguy (changes requested):

app/components/geoblacklight/attribute_table_component.html.erb
thatbudakguy: Locally, I tried a component like:
```erb
<div class="accordion" id="<%= @id %>">
  <div class="accordion-item">
    <h2 class="accordion-header" id="headingOne">
      <button class="accordion-button" type="button" data-bs-toggle="collapse" data-bs-target="#collapseOne" aria-expanded="true" aria-controls="collapseOne">
        <%= @title %>
      </button>
    </h2>
    <div id="collapseOne" class="accordion-collapse collapse show" aria-labelledby="headingOne" data-bs-parent="#<%= @id %>">
      <div class="accordion-body">
        <%= content %>
      </div>
    </div>
  </div>
</div>
```

So that you could then do:
```erb
<% render Geoblacklight::AccordionComponent.new(title:, id: 'index-map-accordion') do %>
  <div id="index-map-attributes" class="viewer-information">
    <em><%= t('geoblacklight.show.click_map') %></em>
  </div>
<% end %>
```

dnoneill: I am seeing something different when they are added to the map. No only does the icon not show up but the header keeps resizing when collapsed. Did you test it with the sidebar?
Screenshot 2025-07-28 at 5 30 52 PM
Screenshot 2025-07-28 at 5 30 48 PM

thatbudakguy:
Everything seems to be working! Only thing I noticed is that when you have `SIDEBAR: false` there's no more border to the attribute table on the bottom, so it looks a little weird (imo):

Screenshot 2025-07-30 at 11 56 04

thatbudakguy (approved):
Works great! We should put a note somewhere (probably https://geoblacklight.org/docs/settings/) about how to enable it.

PR #1468, opened 2024-08-08
move document components to document folder

dnoneill:
Per discussion with Blacklight developer we should be mirroring the stucture of blacklight components with our overrides.


ewlarson (approved):
👍 // Nice nesting here.

PR #1478, opened 2024-08-08
Adjust style paths so Vite aliases aren't needed

thatbudakguy:
This shifts the structure of the stylesheet to separate building
Geoblacklight's styles from the application itself, so it's a
little more clear where things are coming from.

By taking the imports out of the stylesheet in the gem and moving
them into the generated app, we can get rid of some weird-looking
paths and aliases in Vite.


ewlarson (approved):
👍 // This is much cleaner and clearer. Nice work!

Issue #1452, opened 2024-08-01
Externalize JS dependencies when building the frontend


thatbudakguy:
### Descriptive summary

Vite's docs on using "library mode" for production [mention externalizing your dependencies](https://vitejs.dev/guide/build#library-mode), in particular things like React and Vue (React is currently a dependency). I could see us doing this for the bigger stuff, at a minimum, since bundling React on every build of the package makes local dev slow.

Currently, our build sizes are clocking in at several MB:
```
app/assets/javascripts/geoblacklight/geoblacklight.umd.js  2,362.77 kB │ gzip: 728.93 kB │ map: 10,660.30 kB
app/assets/javascripts/geoblacklight/geoblacklight.mjs             0.12 kB │ gzip:   0.13 kB │ map:      0.10 kB
app/assets/javascripts/geoblacklight/raw-BoIBRE52.mjs              0.18 kB │ gzip:   0.17 kB │ map:      0.45 kB
app/assets/javascripts/geoblacklight/deflate-Bu1COhT_.mjs          0.26 kB │ gzip:   0.22 kB │ map:      0.58 kB
app/assets/javascripts/geoblacklight/packbits-CIEu-5mK.mjs         0.59 kB │ gzip:   0.35 kB │ map:      1.53 kB
app/assets/javascripts/geoblacklight/webimage-BYcWplwU.mjs         0.85 kB │ gzip:   0.45 kB │ map:      2.31 kB
app/assets/javascripts/geoblacklight/basedecoder-DHcBySSe.mjs      2.11 kB │ gzip:   0.85 kB │ map:      6.18 kB
app/assets/javascripts/geoblacklight/lzw-4PDRGo2e.mjs              2.12 kB │ gzip:   0.97 kB │ map:      6.93 kB
app/assets/javascripts/geoblacklight/jpeg-OD7BlVnu.mjs            15.25 kB │ gzip:   4.59 kB │ map:     48.20 kB
app/assets/javascripts/geoblacklight/pako.esm-CB1uQYY0.mjs        71.64 kB │ gzip:  17.88 kB │ map:    319.96 kB
app/assets/javascripts/geoblacklight/lerc-Bww1m1Yl.mjs            87.28 kB │ gzip:  29.14 kB │ map:    181.06 kB
app/assets/javascripts/geoblacklight/decoder-DJlmx386.mjs        115.66 kB │ gzip:  43.12 kB │ map:    119.43 kB
[...]

PR #1479, opened 2024-08-08
Revert to using `yarn link` so changes are picked up in development

thatbudakguy:
This fixes an issue introduced with https://github.com/geoblacklight/geoblacklight/pull/1433 where changes to frontend assets (JS, CSS) don't get picked up in local development and require fully rebuilding the `.internal_test_app`.

The problem is that:
- `yarn add` adds the outer assets directory, which ensures that @geoblacklight/frontend gets added with all its dependencies, but doesn't pick up any change thereafter
- `yarn link` links the outer assets directory, which adds @geoblacklight/frontend _without_ its dependencies, but does pick up changes made later

This PR reverts to the latter strategy and adds a note to make sure that we always include all the dependencies of `@geoblacklight/frontend` in the _generated app's_ package.json, not just geoblacklight's package.json, so that they get installed for use in local dev.

Not a perfect solution, but #1452 might improve it further.

### DIFF
--- a/lib/generators/geoblacklight/install_generator.rb
+++ b/lib/generators/geoblacklight/install_generator.rb
@@ -110,19 +110,35 @@ module Geoblacklight
       copy_file "package.json", "package.json"
     end
 
+    # Pick a version of the frontend asset package and install it.
+    def add_frontend
+      # If in local development or CI, install the version we made linkable in
+      # the test app generator. This will make it so changes made in the outer
+      # directory are picked up automatically, like a symlink. Note that this
+      # does NOT install the dependencies of the package, so we have to make
+      # sure to install those separately. See:
+      # https://classic.yarnpkg.com/lang/en/docs/cli/link/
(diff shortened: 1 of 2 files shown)


lib/generators/geoblacklight/install_generator.rb
```diff
@@ -110,21 +110,37 @@ def copy_package_json
...
-    end
-
+    # Pick a version of the frontend asset package and install it.
     def add_frontend
+      # If in local development or CI, install the version we made linkable in
+      # the test app generator. This will make it so changes made in the outer
+      # directory are picked up automatically, like a symlink. Note that this
+      # does NOT install the dependencies of the package, so we have to make
+      # sure to install those separately. See:
+      # https://classic.yarnpkg.com/lang/en/docs/cli/link/
```
ewlarson, line 120: This link is possibly stale, but Yarn 2's docs on link are so much worst than Yarn 1.

ewlarson (approved):
👍 // Appreciate the commentary here.

PR #1485, opened 2024-08-09
Sidebar Map - Display item geom/bbox

### LINKED ISSUE #1350: Map/UX - Adjust the bounding box polygon style on complex geoms
### Descriptive summary

A pending BTAA customization that might be beneficial for GBL proper...

We've customized the bounding box polygon for complex geoms in the BTAA Geoportal to differ from the complex geom layer.

### Original
Screenshot 2024-02-29 at 1 11 49 PM

### Customized
Example 1
Screenshot 2024-02-29 at 11 59 57 AM

Example 2
Screenshot 2024-02-29 at 12 07 19 PM

### Commit

Commit with the customization is here:
https://github.com/geobtaa/geoportal/commit/2ca07815d8079307082dbafc5d2f33a8b81efdc0

### LINKED ISSUE #1482: Bug - Sidebar Map - Missing geom/bbox outlines
### Descriptive summary

The sidebar static map in GBL `main` is missing the geom/bbox outline(s).

### Expected behavior

In GBL v4.4, the sidebar map will render the geom/bbox in the sidebar map:
http://localhost:3000/catalog/p16022coll229:3210

gbl-v4 4-sidebar-map

### Actual behavior

In GBL v5 (main), the sidebar map does not render geom/bbox in the sidebar map:
http://localhost:3001/catalog/p16022coll229:3210

gbl-main-sidebar-map

1. Run v4.4
2. Run a v5 PR
3. Compare sidebar map presentation

### Related work

https://github.com/geoblacklight/geoblacklight/pull/1441

ewlarson:
* Add new BTAA complex-geom.json fixture
* Add leaflet-viewer-draw-initial-bounds-value to static_map_component.rb
* Adjusts logic in the leaflet_viewer_controller.js to support complex geoms
* Styles the bounding box polygon as a dashed border for complex geoms

### Before

Screenshot 2024-08-09 at 3 46 35 PM

### After

Screenshot 2024-08-09 at 3 32 47 PM

Fixes #1482 
Fixes #1350

### DIFF
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -76,7 +76,6 @@ export default class LeafletViewerController extends Controller {
       await this.addPreviewOverlay();
       this.addInspection();
     } else if (this.drawInitialBoundsValue && this.bounds) {
-      if (this.mapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue))
       this.addBoundsOverlay(this.bounds);
     }
 
@@ -166,6 +165,10 @@ export default class LeafletViewerController extends Controller {
       } : undefined
     );
     
+    // Render geometry layer if available
+    if (this.mapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue));
+
+    // Render the bounds overlay
     this.boundsOverlay = boundsOverlay;
     this.overlay.addLayer(boundsOverlay);
   }
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -157,7 +157,7 @@ export default class LeafletViewerController extends Controller {
       bounds.getNorthEast(),
       bounds.getNorthWest(),
     ],
-      this.mapGeomValue ? {
+      hasMapGeomValue ? {
         color: '#3388ff',
         dashArray: "5 5",
         weight: 2,
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -150,6 +150,7 @@ export default class LeafletViewerController extends Controller {
 
   // Add the bounding box to the map
   addBoundsOverlay(bounds) {
+    const hasMapGeomValue = Object.keys(this.mapGeomValue).length > 0;
     const boundsOverlay = polygon([
       bounds.getSouthWest(),
       bounds.getSouthEast(),
@@ -166,7 +167,7 @@ export default class LeafletViewerController extends Controller {
     );
     
     // Render geometry layer if available
-    if (this.mapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue));
+    if (hasMapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue));
 
     // Render the bounds overlay
     this.boundsOverlay = boundsOverlay;
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -156,7 +156,16 @@ export default class LeafletViewerController extends Controller {
       bounds.getSouthEast(),
       bounds.getNorthEast(),
       bounds.getNorthWest(),
-    ]);
+    ],
+      {
+        color: '#3388ff',
+        dashArray: "5 5",
+        weight: 2,
+        opacity: 1,
+        fillOpacity: 0
+      }
+    );
+    
     this.boundsOverlay = boundsOverlay;
     this.overlay.addLayer(boundsOverlay);
   }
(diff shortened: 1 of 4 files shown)


app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -75,7 +75,7 @@ export default class LeafletViewerController extends Controller {
     if (this.availableValue && this.protocolValue) {
       await this.addPreviewOverlay();
       this.addInspection();
-    } else if (this.drawInitialBoundsValue && this.bounds) {
+    } else if ((this.drawInitialBoundsValue || !this.protocolValue) && this.bounds) {
```
dnoneill, line 78: You need one or the other. Either leaflet-viewer-draw-initial-bounds-value or this line. Also I would recommend not doing this. Since you watch a multipolygon, I would change line 75 to if (this.availableValue)
and then add a line after https://github.com/geoblacklight/geoblacklight/blob/0d00b850e58e8ab92f5db97166563bb83afa3ea3/app/javascript/controllers/leaflet_viewer_controller.js#L203, and have the line be
else { return tileLayer(url, options); } which might work. I would have to double check. This definitely is going to need some tinkering.

app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -155,7 +156,16 @@ export default class LeafletViewerController extends Controller {
       bounds.getSouthEast(),
       bounds.getNorthEast(),
       bounds.getNorthWest(),
-    ]);
+    ],
+      {
```
dnoneill, line 160: Do we want this for all views? I am assuming we don't. I would do an if statement here. If mapGeomValue have these settings.

ewlarson: The additional dashed border causes no harm on simple boundaries, as it overlays exactly. Example below:

Screenshot 2024-08-09 at 4 30 44 PM

But, the code would likely be clearer with another conditional... I'll give that a try.

dnoneill: I don't think it needs it if it doesn't apply it to the large area, I'm not quite sure how it isn't though. Maybe I am reading it wrong?

ewlarson: @dnoneill // I made the polyline options conditional here: https://github.com/geoblacklight/geoblacklight/blob/02c677a91f4f6162076da70ca060ebd21894d049/app/javascript/controllers/leaflet_viewer_controller.js#L160-L166

I think that reads a lot better, but let me know what you think!

app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -76,6 +76,7 @@ export default class LeafletViewerController extends Controller {
       await this.addPreviewOverlay();
       this.addInspection();
     } else if (this.drawInitialBoundsValue && this.bounds) {
+      if (this.mapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue))
```
thatbudakguy, line 79: is this different from line 50 because it creates an actual `geoJSON` layer, instead of just returning the bounds of that layer? maybe we should just change line 50 to do that directly, instead of doing it here, since it would work just as well for non-complex items. that avoids the scenario where we have two different layers, one for the bounds (non-complex) and one for the entire geometry (complex).

thatbudakguy: actually, it might be cleaner to just move all this logic to `addBoundsOverlay()`. I think we should only draw one overlay.

thatbudakguy: happy to talk/pair on this to explain what I'm thinking

ewlarson: Sorry! Are you looking to pair with @dnoneill or me? This PR is a bit of a mashup now.

ewlarson: @thatbudakguy // Moved the geom render into addBoundsOverlay. Simplifies that if/else block into calling one function to render added boundary layers. LMKWYT.

thatbudakguy (approved):
looks great; sorry for the holdup! this is exactly what I was thinking.

dnoneill:
Please don't merge. I think this broke the search page.

dnoneill:
It did.  ```if (this.mapGeomValue) this.map.addLayer(L.geoJSON(this.mapGeomValue));``` will run on the search page because this.mapGeomValue equals an empty object.

dnoneill:
@ewlarson see lines https://github.com/geoblacklight/geoblacklight/pull/1505/files#diff-a3e623aa46d009258cd2d99a621af30ef1ff1275d6ed7b330c521418abeca2fdR154 for how to fix.

dnoneill (changes requested):
fix problem with this.mapGeomValue

dnoneill (approved):

PR #1486, opened 2024-08-09
fix complex overlays

dnoneill:


ewlarson (approved):
👏 // Indeed this does it.

PR #1489, opened 2024-08-12
Allow installing with Rails 7.1 and GBL 3.x

eliotjordan:
Rails 6.1 is EOL in October 2024 and Rails 7.0 is EOL April 2025. In order to allow implementers using GBL 3.x more time to upgrade, this PR bumps the max Rails version to 7.1.

https://github.com/rails/rails/pull/52471#issuecomment-2271622034

### DIFF
--- a/lib/geoblacklight/version.rb
+++ b/lib/geoblacklight/version.rb
@@ -1,4 +1,4 @@
 # frozen_string_literal: true
 module Geoblacklight
-  VERSION = '3.7.0'
+  VERSION = '3.8.0'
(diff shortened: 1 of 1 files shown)


jcoyne:
@eliotjordan Rails 7.2 was released. Can we allow that as well?

jcoyne:
I can't see any reason why not. Go for it.

eliotjordan:
@jcoyne https://github.com/projectblacklight/blacklight/pull/3270

PR #1491, opened 2024-08-12
Centralize asset generation and support alternative pipelines

thatbudakguy:
This PR is pre-work for #1467.

It takes some of the asset-related generator functions that were located in other generators and centralizes them in a single, Vite-specific generator. It then borrows an approach from Blacklight and creates a seam for choosing a different asset pipeline setup by selecting a generator at install time.

Vite is kept as the default for now, so this shouldn't result in any changes.


marlo-longley:
Tested locally all works as expected.

PR #1496, opened 2024-08-13
Bump max Rails version to 7.2 in GBL 4

eliotjordan:
Rails 6.1 is EOL in October 2024 and Rails 7.0 is EOL April 2025. In order to allow implementers more flexibility, this PR bumps the max Rails version to 7.2.


jcoyne:
I've hit this inflector error before. Move the `require 'factory_bot'` after the application load. https://github.com/geoblacklight/geoblacklight/blob/main/spec/spec_helper.rb#L23

PR #1497, opened 2024-08-13
Test Rails 7.2 in CI matrix in GBL 5

eliotjordan:
Add Rails 7.2 to CI test matrix.


jcoyne:
@eliotjordan in spec_helper.rb move the `require 'factory_bot'` after the application is initialized by EngineCart. This avoids the missing Inflector error.

thatbudakguy (approved):

PR #1505, opened 2024-08-13
Customizable hover

### LINKED ISSUE #1504: make boundsoverlay customizable
### Descriptive summary
By default the color for bounds overlay is always #3388ff. If you want to update the color you have to go into the leaflet js and update the code. This is annoying and easily fixed by have the settings be set in the settings.yml. Additionally, we can customize to each page (static map, non public, search page).

dnoneill:
closes #1504

### DIFF
--- a/app/javascript/controllers/leaflet_viewer_controller.js
+++ b/app/javascript/controllers/leaflet_viewer_controller.js
@@ -21,7 +21,7 @@ import {
   indexMapLayer,
 } from "../leaflet/layers.js";
 import { geoJSONToBounds } from "../leaflet/utils.js";
-import { DEFAULT_BOUNDS, DEFAULT_OPACITY } from "../leaflet/constants.js";
+import { DEFAULT_BOUNDS, DEFAULT_OPACITY, DEFAULT_GEOM_OVERLAY_OPTIONS } from "../leaflet/constants.js";
 
 export default class LeafletViewerController extends Controller {
   static values = {
@@ -32,6 +32,7 @@ export default class LeafletViewerController extends Controller {
     options: Object,
     basemap: String,
     mapGeom: Object,
+    page: String,
     layerId: String,
     drawInitialBounds: Boolean,
   };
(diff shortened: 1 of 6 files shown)


lib/generators/geoblacklight/templates/settings.yml
```diff
@@ -229,6 +229,13 @@ WMS_PARAMS:
 # Settings for leaflet
 LEAFLET:
   MAP:
+  BOUNDSOVERLAY:
```
jcoyne, line 232: WHY ARE WE SHOUTING?

app/javascript/controllers/leaflet_viewer_controller.js
```diff
@@ -151,19 +152,20 @@ export default class LeafletViewerController extends Controller {
   // Add the bounding box to the map
   addBoundsOverlay(bounds) {
     const hasMapGeomValue = Object.keys(this.mapGeomValue).length > 0;
+    var defaultOptions = hasMapGeomValue ? {
+      color: '#3388ff',
+      dashArray: "5 5",
+      weight: 2,
+      opacity: 1,
+      fillOpacity: 0
+    } : {};
```
thatbudakguy, line 161: could we extract this to a constant (either in this module or in `leaflet/constants.js` called like `DEFAULT_GEOM_OVERLAY_OPTIONS` or something like that?

thatbudakguy: then the line would look like
`const defaultOptions = hasMapGeomValue ? DEFAULT_GEOM_OVERLAY_OPTIONS : {}`

thatbudakguy (approved):

ewlarson:
Rebase and I'll merge.

lib/generators/geoblacklight/templates/settings.yml
jcoyne: Seriously though, how are these values going to get into applications that are being upgraded? This seems like it will only work for newly generated applications.

ewlarson: Yeah, that can be an issue, but `settings.yml` changes will go into our v5 upgrade guide. In the past that has sufficed.

PR #1508, opened 2024-08-13
Refactor the Web Services modal feature tests

### LINKED ISSUE #1480: Flappy test in CI
One of the web services modal tests seems to fail once per build, but usually succeeds when re-running. 
https://github.com/geoblacklight/geoblacklight/blob/4a98fbb4c00bdf26f40d636340382ccab2a1bd6a/spec/features/web_services_modal_spec.rb#L71-L81

See e.g. [attempt 1 of this PR](https://github.com/geoblacklight/geoblacklight/actions/runs/10310312739/attempts/1?pr=1478). The error is:
```
Selenium::WebDriver::Error::ElementClickInterceptedError:
       element click intercepted: Element is not clickable at point (382, 1834)
```

thatbudakguy:
Fixes #1480 ...I think?

This writes a helper method to open the web services modal the same way every time and ensure it has loaded by asserting that its heading is visible.

It also uses more standard rspec syntax (`:feature` test type, describe/expect/it, etc.) and factors out some repeated elements in the tests.

### DIFF
--- a/spec/features/web_services_modal_spec.rb
+++ b/spec/features/web_services_modal_spec.rb
@@ -2,12 +2,33 @@
 
 require "spec_helper"
 
-feature "web services tools" do
-  feature "when wms/wfs are provided", js: true do
-    scenario "shows up in tools" do
+# Open the web services modal and wait for it to load
+def open_web_services_modal
+  expect(page).to have_link "Web services"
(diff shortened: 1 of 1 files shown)


ewlarson (approved):
👍 // Let's give this a try!

PR #1511, opened 2024-08-13
change how citations are shown

### LINKED ISSUE #1500: Adjust citation modal to change how link is created
### Descriptive summary

STUB ticket.  Implementation details here: https://github.com/sul-dlss/earthworks/issues/1168

peetucket:
Removes "Citation info" text from the beginning.

Link shown at the end now varies instead of being a link to the solr document (e.g. earthworks).  Instead, it will look to see if the identifier looks like a URL (could be a DOI or a PURL) and use that if found.  If not found, it will look in the references for a schema.org/url link.  If found, it will use that.  If not, back the current (solr document url)

Fixes #1500

Example:

Screenshot 2024-08-13 at 3 59 04 PM

### DIFF
--- a/config/locales/geoblacklight.en.yml
+++ b/config/locales/geoblacklight.en.yml
@@ -85,7 +85,8 @@ en:
       version_of_descendants: 'Has version...'
       browse_all_no_count: "Browse all records..."
       browse_all: "Browse all %{count} records..."
-
+    modal:
+      copy-citation: "Copy Citation"
     metadata:
       toggle_summary: 'Toggle summary'
       view_metadata: 'View Metadata'
--- a/config/locales/geoblacklight.en.yml
+++ b/config/locales/geoblacklight.en.yml
@@ -2,8 +2,6 @@ en:
   geoblacklight:
     bbox_label: 'Bounding Box'
     location: 'Location'
-    citation:
-      retrieved_from: 'Retrieved from %{document_url}'
     clipboard:
       citation: Citation copied to clipboard
       web_service: Web service copied to clipboard
--- a/app/models/concerns/geoblacklight/solr_document/citation.rb
+++ b/app/models/concerns/geoblacklight/solr_document/citation.rb
@@ -12,11 +12,33 @@ module Geoblacklight
           fetch(Settings.FIELDS.TITLE, nil),
           ("[#{format}]" if format),
           fetch(Settings.FIELDS.PUBLISHER, nil),
-          I18n.t("geoblacklight.citation.retrieved_from", document_url: solr_document_url)
+          citation_link(solr_document_url)
         ].flatten.compact.join(". ")
       end
 
       private
+
+      # The actual citation link varies based on what is available in the document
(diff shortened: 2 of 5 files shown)


config/locales/geoblacklight.en.yml
```diff
@@ -2,8 +2,6 @@ en:
   geoblacklight:
     bbox_label: 'Bounding Box'
     location: 'Location'
-    citation:
```
ewlarson, line 5: Surprised we dropping the "Retrieved from " preface here? — Isn't that APA style? AND... nope, looks like APA 7th drops the "Retrieved from" requirement in citations.

ewlarson (changes requested):
Seeing a couple issues here:

1. Big issue - When I click "Copy Citation", I'm seeing "undefined" when I try to paste the citation
2. Small issue - The dismissable alert button is miss-aligned (screenshot below) — it should float left, when the wrapping alert class element includes "alert-dismissable". See: https://github.com/projectblacklight/blacklight/commit/1bef777b6999e59eb80fa8f65a497e2efbaeade3

Screenshot 2024-08-15 at 11 25 12 AM

Otherwise, these changes look quite close to 🚢

srappel:
While we're here. What do folks think about the idea of adding an APA label to the citation? Could be useful to encourage students to double check their formatting before they drop it into a paper.

Google scholar looks like this:
image

ewlarson:
RE: APA. Let's discuss at sprint stand up today. We might also include a statement indicating these citations might be inchoate.

config/locales/geoblacklight.en.yml
peetucket: Thanks. 

1. Class added to fix close alert alignement.
2. Copy to clipboard was broken due a missing translation (which added a span, which broke the javascript).  Fixed.

ewlarson:
Running the application in development mode, I'm getting this in my clipboard after "Copy Citation":

```
<!-- BEGIN /Users/ewlarson/Rails/geoblacklight/app/components/geoblacklight/document/citation_component.rb -->Cornell Institute for Resource Information Sciences (Cornell IRIS). NYS Department of Agriculture and Markets. (2011-02-28). Test record with additional download formats. [Shapefile]. http://localhost:3000/catalog/cugir-007950<!-- END /Users/ewlarson/Rails/geoblacklight/app/components/geoblacklight/document/citation_component.rb -->
```

Is this just an artifact of dev mode? Is this a ViewComponent thing?

ewlarson:
Please rebase. What's your take on those HTML comments (the component debugging output) in the citation copy value?

peetucket:
> Please rebase. What's your take on those HTML comments (the component debugging output) in the citation copy value?

Rebase done and up to date with main.  I'm not seeing that component debugging output in my clipboard in localhost when I copy and paste a citation, so I can't reproduce it.  I tried a few different local fixtures and they all work for me.

ewlarson (approved):
🚀 // This is ready — we'll keep in mind that annotate_rendered_view_with_filenames development.rb config option should be false.

PR #1513, opened 2024-08-14
Bump frontend package version to 5.0.0-alpha.7

thatbudakguy:
This is the next non-private package number available for publishing (via semver) because I `npm publish`ed the preceding few without doing `--access public` 🤦🏻

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@geoblacklight/frontend",
   "type": "module",
-  "version": "5.0.0-alpha.4",
+  "version": "5.0.0-alpha.7",
   "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


dnoneill:
@thatbudakguy version 3 is still the latest version I see on NPM. I wouldn't bump until that is updated.

thatbudakguy:
@dnoneill see https://geoblacklight.slack.com/archives/C0TCW03EC/p1723666590520069

thatbudakguy:
If I `publish` without a package.json that has that version number in it, the published package contents will not actually reflect the version number they are published under

ewlarson (approved):
🚀

PR #1516, opened 2024-08-14
Publish unbundled versions of JS assets

### LINKED ISSUE #1514: Ship unbundled versions of JS as part of the frontend package
Blacklight does this by simply including `app/javascript` as part of the npm package that it delivers, so the JS is available unbundled and exactly as it is in the gem.

For consumers that aren't going to bundle their JS, this is helpful so that they can include only some parts, or override where the imported libraries like react and leaflet come from via importmaps.

thatbudakguy:
Closes #1514 

This uses the same strategy as blacklight: bundle things into `app/assets/javascripts` but gitignore it, and then ship everything in `app/assets` (including the bundled `app/assets/javascripts`) _and_ `app/javascript` (unbundled).

This way, the paths inside the package match where things go in your built app, and there's no `copyPublicDir` magic needed to make sure that images or other assets in `app/assets` get included in the npm package: they will always have the same path in the package that they do in the gem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@ package-lock.json
 /public/vite*
 /public/geoblacklight-vite*
 node_modules
-dist
+app/assets/javascripts/geoblacklight
 # Vite uses dotenv and suggests to ignore local-only env files. See
 # https://vitejs.dev/guide/env-and-mode.html#env-files
 *.local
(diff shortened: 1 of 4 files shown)


ewlarson (approved):
🏄 // Nice work here!

Issue #1512, opened 2024-08-14
Docker Builds - Vite not installing; Assets missing


ewlarson:
### Descriptive summary

Assets are no longer building correctly in the automated Docker builds in `main` PRs.

### Expected behavior

App has CSS and JS

### Actual behavior

App has no CSS and JS applied.

![Screenshot 2024-08-13 at 10 38 48 AM](https://github.com/user-attachments/assets/73759fc5-b432-4d0d-9f8d-be85ae9b3381)

### Steps to reproduce the behavior

1. Download any recent PR
1. Run the docker commands to spin up the app
2. See the unstyled output.

### Related work

Likely related to recent asset generator work.

ewlarson:
Per @thatbudakguy // Odd networking errors here too:

```
app-1   | Started GET "/assets/application-c2cd8c57.css" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
app-1   | Started GET "/assets/application-bfcdf840.js" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Started GET "/assets/stimulus-loading-1fc53fe7.js" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Started GET "/assets/blacklight/checkbox_submit-152878cd.js" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Started GET "/assets/blacklight/bookmark_toggle-493f5516.js" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
app-1   | Started GET "/assets/turbo.min-29118695.js" for 192.168.65.1 at 2024-08-14 19:18:42 +0000
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
app-1   | Cannot render console from 192.168.65.1! Allowed networks: 127.0.0.0/127.255.255.255, ::1
```

PR #1517, opened 2024-08-15
Bug fix - Use ASSET_PIPELINE in asset generator to fix Docker builds

ewlarson:
Restores Vite as the asset pipeline for Docker builds.

Fixes: #1512

### DIFF
--- a/.github/workflows/build-demo-app.yml
+++ b/.github/workflows/build-demo-app.yml
@@ -35,7 +35,7 @@ jobs:
         run: exec "yarnpkg"
 
       - name: Build
-        run: BRANCH=${{ env.BRANCH }} rails new build -m template.rb -a propshaft
+        run: BRANCH=${{ env.BRANCH }} ASSET_PIPELINE=vite rails new build -m template.rb -a propshaft
 
       - name: Cleanup
         run: rm -rf build/node_modules build/Gemfile.lock build/yarn.lock build/tmp
--- a/.github/workflows/build-demo-app.yml
+++ b/.github/workflows/build-demo-app.yml
@@ -35,7 +35,7 @@ jobs:
         run: exec "yarnpkg"
 
       - name: Build
-        run: BRANCH=${{ env.BRANCH }} ASSET_PIPELINE=vite rails new build -m template.rb -a propshaft
+        run: BRANCH=${{ env.BRANCH }} rails new build -m template.rb -a propshaft
 
       - name: Cleanup
         run: rm -rf build/node_modules build/Gemfile.lock build/yarn.lock build/tmp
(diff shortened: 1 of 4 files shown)


template.rb
```diff
@@ -14,5 +14,6 @@
 
 generate "blacklight:install", "--devise"
 generate "geoblacklight:install", "-f"
+generate "geoblacklight:assets:vite"
```
ewlarson, line 17: Forcing the vite generator here does resolved the assets.

.github/workflows/build-demo-app.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         run: exec "yarnpkg"
 
       - name: Build
-        run: BRANCH=${{ env.BRANCH }} rails new build -m template.rb -a propshaft
+        run: BRANCH=${{ env.BRANCH }} ASSET_PIPELINE=vite rails new build -m template.rb -a propshaft
```
ewlarson, line 38: Adding this env var here did propogate the value through the to docker build.

ewlarson:
Assets via Vite:

Screenshot 2024-08-16 at 3 42 05 PM

thatbudakguy (changes requested):
Vite is the default and shouldn't require any configuration:
https://github.com/geoblacklight/geoblacklight/blob/6776b5ac8c08c041733e76f7f3f4ce6a1d0dd71a/lib/generators/geoblacklight/assets_generator.rb#L7

Is something else messed up here? Is `ASSET_PIPELINE` somehow set to something in the docker container? You shouldn't need to set it at all in order for the vite generator to run. Maybe the `ENV.fetch` logic is somehow wrong?

thatbudakguy:
I wonder if the `class_option` should just default to `vite`, and then we should use `ENV["ASSET_PIPELINE"]` explicitly in the template.rb file?

You should never need to call the asset generator directly; `geoblacklight:install` should invoke it for you.

ewlarson:
> Is something else messed up here? Is ASSET_PIPELINE somehow set to something in the docker container? You shouldn't need to set it at all in order for the vite generator to run. Maybe the ENV.fetch logic is somehow wrong?

Yeah, your code _reads_ fine to me, but for whatever reason, the asset generator just doesn't run... This change is a bit heavy-handed, but fixes the current no-assets Docker builds.

eliotjordan (approved):

eliotjordan:
@ewlarson @thatbudakguy There were a few problems with the install and assets generators - primarily the assets were not generated outside if :test. I fixed that up and was able to get it working without running the assets generator directly in template.rb.

eliotjordan (approved):

Issue #1323, opened 2024-01-10
Remove email downloads from core


dl-maura:
### Descriptive summary

Harvard no longer does email downloads, so the special functionality we built in for Harvard is no longer needed

### Expected behavior

We just don't want this code anymore, so this will make that go away.

PR #1518, opened 2024-08-15
#1323- Remove Email download from core 

yuntianhu:
Harvard no longer does email downloads, so the special functionality we built in for Harvard is no longer needed

Expected behavior
We just don't want this code anymore, so this will make that go away.


app/javascript/geoblacklight/downloaders/downloader.js
```diff
@@ -0,0 +1,60 @@
+export default class GeoBlackLightDownloader {
+    constructor(el, options) {
+        this.el = document.querySelector(el);
```
jcoyne, line 3: Why is this necessary? It looks like you're passing in an element, not a selector.

app/javascript/geoblacklight/downloaders/downloader.js
```diff
@@ -0,0 +1,60 @@
+export default class GeoBlackLightDownloader {
```
jcoyne, line 1: GeoBlacklight is the preferred capitalization.

app/javascript/geoblacklight/downloaders/downloader.js
thatbudakguy: This file can be deleted completely; downloads are handled via the `DownloadController` stimulus controller now.

yuntianhu: seems like we don't need app/javascript/geoblacklight/downloader.js, also?  I will rebase.

thatbudakguy: yep, that's right. rebase should fix it

eliotjordan (approved):

PR #1520, opened 2024-08-15
Rewrite relative imports for importmap compatibility

### LINKED ISSUE #1519: Rewrite relative imports in JS for importmap compatibility
### Descriptive summary

When loading the unbundled JS as part of an importmap, imports like:
```js
import { foo } from './bar/baz.js'
```

...will fail because of the relative path. If you load the main entrypoint via CDN, for example, you will see things like:
> Loading module from “https://cdn.jsdelivr.net/npm/@geoblacklight/frontend@5.0.0-alpha.8/app/javascript/controllers/clover_viewer_controller” was blocked because of a disallowed MIME type (“text/plain”).
[localhost:3000](http://localhost:3000/)
>
> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://cdn.jsdelivr.net/npm/@geoblacklight/frontend@5.0.0-alpha.8/app/javascript/controllers/clover_viewer_controller. (Reason: CORS request did not succeed). Status code: (null).

..because relative references between modules in the JS do not resolve.

Blacklight solves this problem by writing all of its imports in this style:
```js
import { foo } from 'blacklight/bar/baz'
```

...and then [using a rollup plugin](https://github.com/projectblacklight/blacklight/blob/a8b5a3d3a2e8243c8a875d40528eae5c106d83bd/rollup.config.js#L14-L19) to convert them to relative paths during Vite's build, so that the JS can work for either strategy (bundlers and importmaps).

This solution was tested in #1467 and seems to work; Vite uses rollup under the hood so we can use the same plugin Blacklight does.

thatbudakguy:
Closes #1519 

**This PR should have no effect on the generated app.**

This PR makes a few changes to the structure of `app/javascript`:
- everything gets moved under the `geoblacklight/` dir, so all imports start with `geoblacklight/`
- the `GeoblacklightInitializer` is gone (#1503 removed the need for it)
- all the initializers go into their own directory at `geoblacklight/initializers`
- HGL-related JS is removed (it wasn't being referenced in the app at this point anyway)

Then, we change all of the imports like `import { foo } from './bar'` to use `import { foo } from 'geoblacklight/bar'`, which makes them compatible with importmaps. You can use the importmap to make `geoblacklight` resolve to a local file or to a CDN download hosted anywhere.

A rollup plugin is added to Vite to rewrite them back to relative imports when bundling with Vite, which is the same strategy that Blacklight uses.


jcoyne (commented):
👏 well done!

ewlarson (changes requested):
This just needs a rebase to pick up the citation PR's changes and I'll merge.

ewlarson (approved):
👍 // Thanks for all your hard work on our JS this sprint. Much appreciated by all!

marlo-longley:
Very much appreciated and recognized !

Issue #1492, opened 2024-08-12
Fix Fixtures


karenmajewicz:
Broken fixtures by ID:

- [x] 4669301e-b4b2-4c8b-bf40-01b968a2865b - esri dynamic layer-single layer
- [x] purdue-urn-f082acb1-b01e-4a08-9126-fd62a23fd9aa - esri WMS
- [x] mit-f6rqs4ucovjk2 - Polygon shapefile with WMS and WFS
- [x] uva-Norfolk:police_point - Multipoint dataset with WMS and WFS and a colon in the slug and layer ID
- [x] harvard-g7064-s2-1834-k3 - This layer is a georeferenced raster image of the historic paper map
- [x] nyu-test-soil-survey-map - Esri tiled map layer
- [x] cugir-007950 - Point dataset with WMS, WFS, direct download, and FGDC metadata XML and HTML
- [x] nyu-2451-34564 (no CORS)
- [x] nyu_2451_34502 (no CORS)
- [x] tufts-cambridgegrid100-04 (polygon WMS) – doesn't support CORS

karenmajewicz:
A short report on findings here:

- We currently have 46 fixture documents
- There are at least 8 that have broken links of various kinds
- The repeat offenders for link rot have generally been external services from public portals, like Esri, but institutional repos also can be a problem.  For example, right now we have several broken records that were submitted by GeoBlacklight repos that have since closed down or migrated their server.

I have updated the fixtures a few times over the years, fixing links, upgrading them to Aardvark, etc. However, it feels like we are at a point where we have too many docs and possibly redundant tests. Is there another way we could approach testing?  Do we need "real world" examples for everything?  Could docs be reused more often for multiple tests?  

Lots of related tickets and PRs over the years showing the headaches specifically of broken links:

- [ ] #67
- [ ] #238
- [ ] #434 
- [ ] #443
- [ ] #666
- [ ] #704 
- [ ] #843
- [ ] #849 
- [ ] #945 
- [ ] #1144
- [ ] #1145
- [ ] #1183 
- [ ] #1408 
- [ ] #1526

thatbudakguy:
Now that NYU has turned on CORS on their geoserver, all of these have been fixed, and all of our fixtures should resolve to something previewable. That's no guarantee they'll stay, but thanks to the test setup, at least that shouldn't break any tests any more.

PR #1526, opened 2024-08-15
Change tests to stop using NYU fixture

thatbudakguy:
This fixes an HTTP timeout error that could arise when an NYU
fixture was queried over WMS and didn't respond in time, causing
tests to fail.

This commit switches the two tests that used the fixture to use
other, equivalent fixtures to avoid the timeouts.

See also #1492.


ewlarson (approved):
👍

PR #1527, opened 2024-08-15
fix file_format to not require the field

dnoneill:


thatbudakguy (approved):

app/models/concerns/geoblacklight/solr_document.rb
```diff
@@ -88,7 +88,7 @@ def wxs_identifier
     end
 
     def file_format
-      fetch(Settings.FIELDS.FORMAT)
+      fetch(Settings.FIELDS.FORMAT, "")
```
ewlarson, line 91: Looks like this could poise an issue a few places:
* https://github.com/geoblacklight/geoblacklight/blob/main/app/components/geoblacklight/download_links_component.html.erb#L16
* https://github.com/geoblacklight/geoblacklight/blob/main/app/components/geoblacklight/download_links_component.rb#L77

Format is also an odd field as it is conditionally required in Aardvark:
https://opengeometadata.org/ogm-aardvark/#format

What's the case for making this non-required?

thatbudakguy: I think this arises from a situation where we wanted to be able to call `file_format` to access the value in `dct_format_s` (if there was one) in a different context, and have it not error.

Since the method is public, it seemed like it could conceivably be called by downstream apps that might not be expecting that it could error (since `dct_format_s` can legitimately be empty) instead of just returning `nil` or an empty string.

If you provide a download link but don't add a `dct_format_s`, the app will blow up when you visit that record's page, which isn't ideal...the record is technically invalid, but I feel like we could have better behavior than just a 500 error.

thatbudakguy: Maybe `file_format` should return `nil` if there's no `dct_format_s`, and then the places it's called should be updated to handle that `nil` and not render a download link.

ewlarson: I like this last option the best.

ewlarson (changes requested):
Let's run with Nick's last thought here: "Maybe file_format should return nil if there's no dct_format_s, and then the places it's called should be updated to handle that nil and not render a download link (and maybe warn?)."

dnoneill:
closing in favor of https://github.com/geoblacklight/geoblacklight/commit/ee6c445cc72a0939634251fbda857031ab1bb3ce#diff-ecd3aa94f057dad57dd45c84020ac9298727f0f2b840c22585cb217b5c036d3aR22

PR #1533, opened 2024-08-16
Remove unused helper method: show_help_text?

jcoyne:


eliotjordan (changes requested):
The next (related) helper function can be removed as well.
`render_help_text_entry`

Otherwise this looks good.

eliotjordan (approved):

PR #1536, opened 2024-08-16
Use zeitwerk

jcoyne:
Zeitwerk is the autoloader for rails.  We can use it and then we don't have to have so many "require" statements.


eliotjordan (changes requested):
@jcoyne This looks good to me. Can you add some text to the the PR comments about what this is and why this is a good idea?

jcoyne:
@eliotjordan done.

eliotjordan (approved):

PR #1549, opened 2024-08-17
Start docker rails server on port 3001

eliotjordan:
Reduce confusion when running the demo app docker. The rails server log now matches the port that compose binds to.


ewlarson (approved):
👍

PR #1551, opened 2024-08-19
Use docker geoserver

eliotjordan:
[WIP] Experiment to use pre-built docker geoserver container.


thatbudakguy:
Going to close this since it's drifted pretty far from `main` (many of the fixtures have changed) and we've settled on a different approach to avoiding test flakiness for now: just don't wait for HTTP requests to external services (e.g. #1842)

PR #1555, opened 2024-08-22
add click to wake to map

### LINKED ISSUE #1554: add click to wake to core geoblacklight
Currently big ten does click to wake on their maps. Add this to core geoblacklight.

dnoneill:
closes #1554

Default behavior

https://github.com/user-attachments/assets/559f0079-b8ec-42e0-8365-3a5a25cc4a65

Settings updated to include hover/new message
Screenshot 2024-08-22 at 2 28 44 PM
Screenshot 2024-08-22 at 2 27 33 PM

### DIFF
--- a/app/javascript/geoblacklight/leaflet/controls/sleep.js
+++ b/app/javascript/geoblacklight/leaflet/controls/sleep.js
@@ -1,0 +1,100 @@
+import { DomUtil, DomEvent, Handler } from "leaflet";
+
+export default class Sleep extends Handler {
+  addHooks() {
+    const container = DomUtil.create("div", "wake-overlay w-100 h-100 position-absolute align-items-center justify-content-center", this._map._container);
+    const button = DomUtil.create("button", "btn btn-primary wake-button", container)
+
+    button.innerHTML = this._map.options.MESSAGE;
+    button.setAttribute("tabindex", "0");
+    button.style.zIndex = 2000;
+    container.style.zIndex = 1999;
+    const background = String(this._map.options.BACKGROUND);
+    container.style.background = background;
+    this.container = container;
+    if (this._map.options.MARGIN_DISTANCE) {
+      this.updateBasedOnSize()
+    } else {
+      this.sleepMap();
+    }
+    this.setListeners()
+  }
+
+  sleepMap() {
+    this._map.dragging.disable();
+    this._map.touchZoom.disable();
+    this._map.scrollWheelZoom.disable();
+    this._map.doubleClickZoom.disable();
+    this._map.boxZoom.disable();
+    this._map.keyboard.disable();  
+    this.container.classList.add('d-flex');
+    this.container.classList.remove('d-none');
+  }
+
+  wakeMap() {
+    this._map.dragging.enable();
+    this._map.touchZoom.enable();
+    this._map.scrollWheelZoom.enable();
+    this._map.doubleClickZoom.enable();
+    this._map.boxZoom.enable();
+    this._map.keyboard.enable();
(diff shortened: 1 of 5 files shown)


app/javascript/geoblacklight/leaflet/controls/sleep.js
```diff
@@ -0,0 +1,75 @@
+import { DomUtil, DomEvent, Handler } from "leaflet";
+
+export default class Sleep extends Handler {
+  addHooks() {
+    const container = DomUtil.create("div", "d-none wake-overlay w-100 h-100 d-flex align-items-center justify-content-center", this._map._container);
+    const button = DomUtil.create("button", "btn btn-primary wake-button", container)
+
+    console.log(this._map.options)
```
marlo-longley, line 8: console!

app/javascript/geoblacklight/leaflet/controls/sleep.js
```diff
@@ -0,0 +1,78 @@
...
+
+    button.innerHTML = this._map.options.MESSAGE;
+    button.setAttribute("tabindex", "0");
+    button.style.zIndex = 2000;
+    container.style.zIndex = 1999;
+    const background = String(this._map.options.BACKGROUND);
+    container.style.background = background;
+    this.container = container;
+    this.sleepMap();
+    this.setListeners(this._map, container)
```
marlo-longley, line 16: in one case you pass `this._map` and in the other you don't, and keep calling it (lines 20-25). Maybe use the same way in both methods?

dnoneill: Good catch. This was cruft from another way I was looking at doing it.

eliotjordan (changes requested):
Let's wait on this until we've had some community input.

dnoneill:
I am fine with that, but by default it doesn't show up. It only shows up if you enable it in settings. https://github.com/geoblacklight/geoblacklight/pull/1555/files#diff-0bdd5132c39335d653c919e990b71334aae05839ee2705141c1d0e8f4d0beb79R240-R246

dl-maura:
+1 to having this since it is configurable -- those of us who have customized our sites to have a larger map view rely on it so people scrolling don't get 'trapped', and would most likely back port it in. Since it's default off and you only enable it if you want it, I don't see it negatively impacting core users unless something about the JS required is EoL

thatbudakguy (approved):
Just tested the resize-detection feature and it seems to work as expected. I had thought that the preferred way to implement this was [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), but evidently [the W3C says that you should just use the `resize` event unless you can't](https://w3ctag.github.io/design-principles/#events-vs-observers), and any performance differences are probably not relevant here.

I think this is ready.

eliotjordan (commented):
It's been a while since I looked at this and I may have missed some conversation, but this is now set to be enabled by default. Is this intended? Also, when enabled, the click to wake functionality only appears to be active at smaller (mobule) screen widths. Is that intended as well?

thatbudakguy:
> It's been a while since I looked at this and I may have missed some conversation, but this is now set to be enabled by default. Is this intended? Also, when enabled, the click to wake functionality only appears to be active at smaller (mobule) screen widths. Is that intended as well?

Both are intentional, I think. The logic calculates whether the map will take up almost all of the screen width (and thus present a scroll trap), so it should only display in cases where that's true – and I can't think of a reason you'd want to disable it in those cases.

eliotjordan (approved):

PR #1563, opened 2024-08-23
Configure the logo rather than overriding

jcoyne:
This makes it easier for downstream apps to configure their own logos and results in less overall css.

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/_blacklight_overrides.scss
@@ -2,15 +2,6 @@
   Stylesheet for overriding styles inherited from Blacklight proper
 **/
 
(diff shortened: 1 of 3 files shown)


eliotjordan (changes requested):
Though it seems like this should work, the generated app has a blank logo.
Screenshot 2024-08-23 at 11 27 27 AM

jcoyne:
Screenshot 2024-08-23 at 12 02 48 PM

jcoyne:
This has got to be broken on the main branch too, right?  Because image_url is a sprockets thing and we're using propshaft?

jcoyne:
Nope, in main:
Screenshot 2024-08-23 at 12 07 39 PM

jcoyne:
@eliotjordan I had to update the vite config for the so that It can deal with the `yarn link` that we do in the test app generator

eliotjordan (approved):

PR #1575, opened 2024-08-27
Revert "Prune unused styles"

eliotjordan:
Reverts geoblacklight/geoblacklight#1562

Some viewers do not display without min height.

### DIFF
--- a/app/assets/stylesheets/geoblacklight/modules/item.scss
+++ b/app/assets/stylesheets/geoblacklight/modules/item.scss
@@ -1,8 +1,30 @@
 #document {
   .document-metadata.dl-invert {
-    dt {
+     dt {
       font-weight: bold;
     }
+  }
+
+  #viewer-container {
+    .viewer {
+      border: 1px solid $gray-400;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/geoblacklight/modules/item.scss
```diff
@@ -1,10 +1,32 @@
 #document {
   .document-metadata.dl-invert {
-    dt {
+     dt {
       font-weight: bold;
     }
   }
 
+  #viewer-container {
```
jcoyne, line 8: My assertion when removing this was that `#viewer-container` is not a child of `#document`. Is that wrong?

eliotjordan: Yeah. It's a child of #document.

eliotjordan: document -> document-main-section -> row -> viewer-container

jcoyne (approved):

PR #1583, opened 2024-08-30
Csvdownload

### LINKED ISSUE #1584: Add CSV option as download option for vector data and add ability to configure which formats are listed for export
### Descriptive summary

Enable a new option for downloads for vector data: CSV downloads.  Enable the definition of which download formats will be listed for export options in the Settings file, enabling different institutions to configure which options they would like to display.  Allow for the current list to be in the default Settings file as well as in the default code.

hudajkhan:
What this pull request does:
Adds CSV download option and functionality
Adds a method to configure which vector download formats for shapefiles we would like to include in the display, with a default list available both in the generated template as well as the code

Closes #1584


app/components/geoblacklight/download_links_component.html.erb
```diff
@@ -22,6 +22,7 @@
             <% end %>
 
             <% if document.download_types.present? %>
+            <%= document.download_types.to_s %>
```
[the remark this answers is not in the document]
hudajkhan: indeed :)

lib/generators/geoblacklight/templates/settings.yml
```diff
@@ -216,6 +216,13 @@ RELATIONSHIPS_SHOWN:
     label: geoblacklight.relations.version_of_descendants
     query_type: descendants
 
+# Vector data download formats to be displayed
+DOWNLOAD_FORMATS:
+  VECTOR:
+    - 'Shapefile'
+    - 'KMZ'
+    - 'GeoJSON'
```
eliotjordan, line 224: I don't see a problem with making CSV download a standard option. I like that the download formats are configurable, but adding CSV to this list make it's easy to test and develop. Better not to hide features by default.

eliotjordan (changes requested):
See comment in settings.yml. Otherwise, this is looking good.

lib/geoblacklight/references.rb
```diff
@@ -72,9 +73,9 @@ def download_types
     ##
     # Returns all of the Esri webservices for given set of references
     def esri_webservices
-      %w[tiled_map_layer dynamic_map_layer feature_layer image_map_layer].map do |layer_type|
+      %w[tiled_map_layer dynamic_map_layer feature_layer image_map_layer].filter_map do |layer_type|
```
eliotjordan, line 76: comment: I like `filter_map`! Much better than the `compact` at the end.

lib/generators/geoblacklight/templates/settings.yml
hudajkhan: Will do!

lib/geoblacklight/references.rb
hudajkhan: I can't take credit for this because I am pretty sure this was a rubocop automatic fix, but good to know it is better!

eliotjordan (approved):

hudajkhan:
thank you!

PR #1609, opened 2024-09-12
Update resource type icon logic

thatbudakguy:
This slightly refactors the logic in the header icons component that picks a more specific icon to display based on an item's resource type.

For cases where an item had multiple resource types, but one of them was one that mapped to an icon (e.g. "Raster data" and "Nautical charts"), we were sometimes falling back to the resource class ("Datasets"). This change has us use "Raster data" since we know we have a more specific icon for it in that situation.

For more info, see https://github.com/sul-dlss/earthworks/issues/1398.

### DIFF
--- a/app/components/geoblacklight/header_icons_component.rb
+++ b/app/components/geoblacklight/header_icons_component.rb
@@ -10,17 +10,20 @@ module Geoblacklight
       super
     end
 
-    def get_icon(field)
-      icon_name = @document[field]
-      if icon_name&.include?("Datasets") && @document.resource_type
-        specific_icon = @document.resource_type
-        specific_icon = specific_icon.is_a?(Array) ? specific_icon.first : specific_icon
-        specific_icon = specific_icon&.gsub(" data", "")
-        icon = geoblacklight_icon(specific_icon, classes: "svg_tooltip")
-        return icon unless icon.include?("icon-missing")
-      end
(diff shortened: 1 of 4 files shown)


app/components/geoblacklight/header_icons_component.rb
```diff
@@ -11,16 +11,19 @@ def initialize(document:, fields: [Settings.FIELDS.RESOURCE_CLASS, Settings.FIEL
...
-      icon_name = icon_name.is_a?(Array) ? icon_name.first : icon_name
-      geoblacklight_icon(icon_name, classes: "svg_tooltip")
+      return get_resource_icon if field == Settings.FIELDS.RESOURCE_CLASS
+
+      geoblacklight_icon(@document[field], classes: "svg_tooltip")
+    end
+
+    # If the item has a resource type of "X data" where "X" is an existing icon, use that icon
+    # Otherwise just use the resource class icon
+    def get_resource_icon
```
jcoyne, line 21: `get_` as a prefix on an accessor method is not typical in Ruby.  https://github.com/rubocop/ruby-style-guide#accessormutator-method-names  Would `resource_icon` work?

jcoyne: To be clear, the other method was not an accessor, so doesn't really fall under the same convention, but I still like to avoid it.

jcoyne (approved):

PR #1612, opened 2024-09-20
Remove solr_wrapper dependency and use docker

### LINKED ISSUE #1611: Replace solr_wrapper with docker compose
Fully remove solr_wrapper and use compose for test and development.

eliotjordan:
Closes #1611

### DIFF
--- a/.github/workflows/build-demo-app.yml
+++ b/.github/workflows/build-demo-app.yml
@@ -43,8 +43,8 @@ jobs:
       - name: Copy Dockerfile
         run: cp lib/generators/geoblacklight/templates/demo-app/Dockerfile build/.
 
-      - name: Copy compose.yml
-        run: cp lib/generators/geoblacklight/templates/demo-app/compose.yml build/.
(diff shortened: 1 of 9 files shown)


thatbudakguy:
I might be missing something, but does running `bundle exec rake ci` work for you? For me, the Docker stack is not started, and all the tests that need to connect to solr fail. There's no output in the terminal indicating that the `system` call to invoke compose happened.

If this is indeed a bug, I think it might be being masked in CI by the fact that the CI build already invokes compose, so it will have the containers running regardless:
https://github.com/geoblacklight/geoblacklight/blob/e7a283161e57d21ab543fdf13f9fec0737d288a8/.github/workflows/ruby.yml#L59-L60

eliotjordan:
@thatbudakguy Good catch. Updated the PR to start and stop solr as part of the ci task.

thatbudakguy (approved):
running tests went smoothly this time; everything looks good.

PR #1635, opened 2025-01-08
Update rails versions

jcoyne:


thatbudakguy:
became part of #1636.

PR #1648, opened 2025-02-18
Use the publicly available seed rake task for generated apps

thatbudakguy:
This fixes an issue where apps newly generated using the
template.rb can't start because they can't access the internal
version of the seed task, which is defined in the Rakefile
instead of in lib/.

We want the bundled version of the seed task, not the top-level
one.


thatbudakguy:
Fixes #1668?

PR #1691, opened 2025-11-13
Change bootstrap imports in initializers

thatbudakguy:
This fixes an error in the importmap build because the Tooltip
and Popover classes weren't export from the bootstrap module
at the top level.


dl-maura (approved):
worked locally for me

PR #1695, opened 2025-11-14
Add weekly dependency updates with Dependabot

thatbudakguy:
Adds config for ruby, js, and github actions updates.


thatbudakguy:
I thought dependabot could also handle updates to `gemspec`?

jcoyne (approved):

Issue #1712, opened 2025-12-19
Bring back docker builds?


jcoyne:
> ERROR: failed to build: failed to solve: failed to copy: httpReadSeeker: failed open: content at https://ghcr.io/v2/geoblacklight/geoblacklight/manifests/sha256:f95563d956b0a156981ceba7f66630bdebeb5f6d2be860962624e9330ee53d82 not found: not found

jcoyne:
Is there a reason we want them?

thatbudakguy:
My understanding is that they were used as a way to preview changes in PRs, but I'm not the one who set them up and I didn't ever use them, so maybe someone else can speak to that. This issue is here in case someone wants to pick the thread back up.

thatbudakguy:
Closing this; we can revisit if someone expresses a need for docker as a way to review PRs.

PR #1717, opened 2026-01-06
Bump the ruby version in demo-app docker

eliotjordan:
Work on https://github.com/geoblacklight/geoblacklight/issues/1712


thatbudakguy:
n.b. https://github.com/geoblacklight/geoblacklight/pull/1690 removed these builds until we can get them passing again. to work on this issue, one could start by reverting that commit and then continuing to assess the failures.

Issue #1627, opened 2024-12-02
Discrepancy between Aardvark Documentation and JSON Schema file


srappel:
https://github.com/geoblacklight/geoblacklight/blob/36fa3977309891a026baf43516bff7bc7f4a6803/schema/geoblacklight-schema-aardvark.json#L193-L195

In the OGM documentation, [suppressed](https://opengeometadata.org/ogm-aardvark/#suppressed) and [georeferenced](https://opengeometadata.org/ogm-aardvark/#georeferenced) are both listed as Field Type **Boolean or String**.

However, the above lines in the JSON schema for Aardvark  only allow for Boolean so string values fail validation.

This seems separate from this (closed) issue, but worth referencing just in case: https://github.com/geoblacklight/geoblacklight/issues/1262

PR #1745, opened 2026-01-30
Fix bug in Aardvark JSON validation file

rmseifried:
Addresses https://github.com/geoblacklight/geoblacklight/issues/1627

The Aardvark schema validation file only permits boolean values for `gbl_suppressed_b` and `gbl_georeferenced_b`, but the metadata schema allows both string and boolean values. This PR adds "string" as an option for validation.


thatbudakguy (approved):
IMO the spec should only allow boolean, but at least this makes us compliant with the spec!

PR #1769, opened 2026-05-28
Remove institutional icons

### LINKED ISSUE #1760: Remove institutional icons
Maintaining an SVG icon for every participating OGM institution causes a lot of issues:
- We have to find or create a new SVG, make a PR creating a new view component, and release a new GBL just to add one institution
- If someone indexes Aardvark records with a provider field that doesn't map to an icon, it just mysteriously has no icon in the UI
- Some of the institutions have logos that don't read well or translate to icons

I propose we do away with icons for each institution entirely, removing the need to maintain them.

thatbudakguy:
This removes all of the icons that correspond to specific
institutions.

Also removes the settings lookup for icon names, which was
used to map institution names to shorter versions, and some
helper methods for rendering the institutional icons in
facets or on the page.

Closes #1760


jcoyne (approved):

thatbudakguy:
## show page
Screenshot 2026-05-29 at 09 00 07

## search results
Screenshot 2026-05-29 at 09 00 01

## provider facet
Screenshot 2026-05-29 at 09 00 53

PR #1777, opened 2026-06-02
Convert styles to CSS; remove frontend package

### LINKED ISSUE #1734: Convert all SCSS into pure CSS
Now that Blacklight 9 has done this, we can ditch our SCSS and no longer have a need to compile to CSS. This should allow us to get rid of the frontend package entirely and just deliver all assets via the gem.

thatbudakguy:
This PR removes all of the existing SCSS stylesheet files and attempts to build up a minimal stylesheet using pure CSS only. A new file `customizations.css` is generated for each built Geoblacklight app, where users can put whatever CSS customizations they would like.

With this approach, gems like cssbundling-rails are not needed, and there is no compilation step required for the frontend files. Everything is delivered directly via propshaft/importmaps, so there is no frontend package to publish or install.

npm and yarn are also no longer needed, since nothing is symlinked and no build step needs to happen for the javascript.


jcoyne:
Is this a major version change?

thatbudakguy:
Yes, it's part of v6 (current `main`).

lib/geoblacklight/engine.rb
```diff
@@ -18,21 +18,22 @@ class Engine < ::Rails::Engine
...
     end
 
-    PRECOMPILE_ASSETS = %w[favicon.ico geoblacklight/geoblacklight.umd.cjs geoblacklight/geoblacklight.umd.cjs.map geoblacklight/geoblacklight.js geoblacklight/geoblacklight.js.map].freeze
+    PRECOMPILE_ASSETS = %w[favicon.ico].freeze
 
     # Make the source JS and CSS available to built apps
     initializer "geoblacklight.assets" do |app|
-      app.config.assets.paths << Geoblacklight::Engine.root.join("app/javascript")
-      app.config.assets.paths << Geoblacklight::Engine.root.join("app/assets")
+      app.config.assets.paths << Geoblacklight::Engine.root.join("app/assets/javascript")
```
jcoyne: Isn't app/javascript the default location in rails?  Why have you moved this path?

thatbudakguy: I didn't realize that – I thought we had moved it outside of app/assets as a way to distinguish built artifacts from source files when doing the vite build.

jcoyne (approved):

lib/geoblacklight/engine.rb
thatbudakguy: Updated.

lib/geoblacklight/engine.rb
```diff
@@ -18,21 +18,22 @@ class Engine < ::Rails::Engine
...
 
-    PRECOMPILE_ASSETS = %w[favicon.ico geoblacklight/geoblacklight.umd.cjs geoblacklight/geoblacklight.umd.cjs.map geoblacklight/geoblacklight.js geoblacklight/geoblacklight.js.map].freeze
+    PRECOMPILE_ASSETS = %w[favicon.ico].freeze
 
     # Make the source JS and CSS available to built apps
     initializer "geoblacklight.assets" do |app|
       app.config.assets.paths << Geoblacklight::Engine.root.join("app/javascript")
-      app.config.assets.paths << Geoblacklight::Engine.root.join("app/assets")
+      app.config.assets.paths << Geoblacklight::Engine.root.join("app/assets/images")
+      app.config.assets.paths << Geoblacklight::Engine.root.join("app/assets/stylesheets")
```
jcoyne, line 27: Fix this

jcoyne (changes requested):

jcoyne (approved):

PR #1780, opened 2026-06-03
Create a geoblacklight configuration with a well defined API

### LINKED ISSUE #1776: Support lowercased config values in settings.yml
This is the default used by `ruby-config`; it [transforms uppercase envvars to lowercase by default](https://github.com/rubyconfig/config#fine-tuning).

jcoyne:
Stop relying on a generated configuration in the application.  This just adds the first step of the configuration. If this is merged we can follow on with other parts of the config.  When complete, GBL should be able to run without requiring a whole copy of the default config in the generated app.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,9 @@ gemspec path: File.expand_path(".")
 # engine_cart: 2.6.0
 # engine_cart stanza: 2.5.0
 # the below comes from engine_cart, a gem used to test this Rails engine gem in the context of a Rails app.
-file = File.expand_path("Gemfile", ENV["ENGINE_CART_DESTINATION"] || ENV["RAILS_ROOT"] || File.expand_path(".internal_test_app", File.dirname(__FILE__)))
+file = File.expand_path("Gemfile",
+  ENV["ENGINE_CART_DESTINATION"] || ENV["RAILS_ROOT"] || File.expand_path(".internal_test_app",
+    File.dirname(__FILE__)))
 if File.exist?(file)
   begin
     eval_gemfile file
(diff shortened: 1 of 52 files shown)


thatbudakguy:
Using `settings.yml` for configuration (and having it populated when the app is generated) is the pattern I'm used to in Rails apps. That said, it seems like Blacklight doesn't do this anymore (if it ever used to).

Can you say more about what this accomplishes? And how the generated app would go about overriding settings, if not using `settings.yml`?

jcoyne:
@thatbudakguy you local app is free to use Settings if you want to.  But we don't need this dependency forced on us by GBL. Furthermore, without a well defined interface, we have no idea of which properties are in use (e.g. what's deprecated, what is removed).  That's caused me a lot of confusion of late.  Blacklight itself has never used Settings.

thatbudakguy:
> Furthermore, without a well defined interface, we have no idea of which properties are in use (e.g. what's deprecated, what is removed). That's caused me a lot of confusion of late.

This part makes sense to me – it's easy to have outdated settings that do nothing linger in `settings.yml`, and also we forget to update the template `settings.yml` in Geoblacklight.

How do you go about changing a setting using this strategy? Do you override the `lib/geoblacklight/configuration` files in your own app?

jcoyne:
@thatbudakguy right now nothing changes. We read in the Settings file. However, we can significantly pair down the settings file to only things that are not the default.

thatbudakguy:
OK, I think I see – this is creating a way for GBL to maintain its own defaults for all the relevant configuration, without needing to have a generated `settings.yml`. That way the file in the app is only for things you've changed from the defaults.

thatbudakguy (approved):

PR #1781, opened 2026-06-03
Provide gemspec path to prevent bundler clash

jcoyne:
This prevents the following error:
```
[!] There was an error parsing `Gemfile`: You cannot specify the same gem twice coming from different sources.
You specified that geoblacklight (>= 0) should come from gemspec at `.` and gemspec at `/Users/jcoyne85/workspace/geoblacklight/geoblacklight`
. Bundler cannot continue.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"
 
 # Specify your gem's dependencies in geoblacklight.gemspec
-gemspec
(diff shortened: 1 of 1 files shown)


thatbudakguy:
I always wondered where this warning came from!!

thatbudakguy (approved):

PR #1782, opened 2026-06-03
Deprecate USE_GEOM_FOR_RELATIONS_ICON

jcoyne:


thatbudakguy:
Shouldn't this PR be against the `release-5.x` branch, and then we can outright remove it in `main` (v6?)

PR #1783, opened 2026-06-03
Deprecate USE_GEOM_FOR_RELATIONS_ICON

jcoyne:


thatbudakguy:
`GEOM_TYPE` is a 1.0 schema field that's not in Aardvark.

PR #1787, opened 2026-06-03
Remove unused key

jcoyne:


thatbudakguy:
Is this actually removing anything, or just reformatting the yml? I'm fine with the latter, but the PR title seems misleading

lib/generators/geoblacklight/templates/settings.yml
```diff
@@ -213,40 +213,39 @@ RELATIONSHIPS_SHOWN:
...
+  :VERSION: "1.1.1"
+  :REQUEST: "GetFeatureInfo"
+  :STYLES: ""
+  :SRS: "EPSG:4326"
+  :EXCEPTIONS: "application/json"
+  :INFO_FORMAT: "application/json"
 
 # Settings for leaflet
 LEAFLET:
-  MAP:
```
jcoyne, line 233: This line is removed.

thatbudakguy: Can we separate that out from the YAML reformatting, so the diff is clearer?

thatbudakguy (approved):

PR #1790, opened 2026-06-04
Add prettier to format javascript

jcoyne:


.github/workflows/ruby.yml
```diff
@@ -20,6 +20,19 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Node
+        uses: actions/setup-node@v5
+        with:
+          node-version: "24"
+      - name: Install dependencies
+        run: yarn install
+      - name: Run Prettier check
+        run: yarn lint
```
thatbudakguy:
```suggestion
      - name: Install dependencies
        run: npm install
      - name: Run Prettier check
        run: npm run lint
```

I know yarn has been used here in the past, but can we use npm? I don't think there is any reason to prefer yarn.

thatbudakguy (approved):

PR #1792, opened 2026-06-04
Format json and yaml files with prettier too

jcoyne:


thatbudakguy (approved):
Might want to change the name of the CI job/task now, since it's not just JS linting anymore.

PR #1793, opened 2026-06-04
Add puffing-billy to cache front end resources

jcoyne:


thatbudakguy:
Closing since I think we determined this path was too much work – the current approach (don't wait on network calls to external services or write tests reliant on them) seems to be working OK for now.

PR #1795, opened 2026-06-08
Relax rails dependency

eliotjordan:
Relax Rails dependency to allow version 8.1.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       matrix:
         ruby_version: ['3.2', '3.3', '3.4']
-        rails_version: ['7.1.5.1', '7.2.2.1', '8.0.2']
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,7 +30,9 @@
     "@samvera/clover-iiif": "^2.3.2",
     "ol": "8.1.0",
     "ol-pmtiles": "^0.3.0",
-    "react": "^18.2.0"
+    "react": "^18.2.0",
+    "react-dom": "^19.2.7",
+    "swiper": "^12.2.0"
```
jcoyne: Is this needed for rails 8.1?

jcoyne (changes requested):

package.json
eliotjordan: We're trying to fix an issue with Clover.

thatbudakguy (approved):

PR #1800, opened 2026-06-11
Remove old schema

jcoyne:


thatbudakguy:
I think we need to confirm first that this is stored somewhere in the OpenGeoMetadata github org before removing it. If this is the only copy of this schema, we need to preserve it (although the GBL repo isn't the place).

thatbudakguy:
I've compared with the one at https://raw.githubusercontent.com/geoblacklight/geoblacklight/refs/heads/main/schema/geoblacklight-schema-1.0.json and it seems like the only differences are whitespace and some links in docs, so let's go ahead and remove this – there is an authoritative copy elsewhere.

thatbudakguy (approved):

PR #1815, opened 2026-06-16
Move the record title to the top of the page

### LINKED ISSUE #1773: Move record title above viewer
This helps it not get pushed too far down when the attribute table is open, and is more consistent with the usual catalog-type behavior.

Image

thatbudakguy:
Also removes the icons from the record title on the show page.

This helps create a more consistent UI so that the title isn't
pushed down the page by the viewer and attribute table. The icons
aren't as necessary on the show page and often wrap awkwardly.

Closes #1773

### DIFF
--- a/app/assets/stylesheets/geoblacklight/record.css
+++ b/app/assets/stylesheets/geoblacklight/record.css
@@ -1,11 +1,15 @@
 /* Record show page styles */
 
 /* Document header */
-.blacklight-catalog-show .documentHeader h1 > * {
+.documentHeader h1 {
+  margin-bottom: 0.5em;
+}
+
+.documentHeader h1 > * {
   vertical-align: middle;
 }
 
-.blacklight-catalog-show .documentHeader .blacklight-icons {
+.documentHeader .blacklight-icons {
   display: inline-flex;
   justify-content: center;
   align-items: center;
@@ -14,7 +18,12 @@
 
 /* Attribute table */
 [data-sidebar=false] .attribute-data {
-  margin: 1rem 0;
+  margin-top: 1rem;
+}
+
+/* Metadata */
+.document-metadata {
+  margin-top: 1rem;
 }
 
 /* Sidebar */
--- a/app/components/geoblacklight/document_component.html.erb
+++ b/app/components/geoblacklight/document_component.html.erb
@@ -12,6 +12,7 @@
     <%= body %>
   <% else %>
     <div class="document-main-section">
+      <%= title %>
       <% @document.display_note.each do |display_note| %>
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/geoblacklight/record.css
```diff
@@ -1,6 +1,10 @@
 /* Record show page styles */
 
 /* Document header */
+.blacklight-catalog-show .documentHeader h1 {
+  margin-bottom: 0.5em;
```
jcoyne, line 5: any reason for mixing `em` and `rem` units?

thatbudakguy: if it's spacing around a text element, I use em so it's proportional to the text size.

I use rem for pretty much everything else.

jcoyne: Is it necessary to set our own margin-bottom here that overrides the margin-bottom set by bootstrap (`0.5rem`)?  This seems like a local choice (per app) and we're making it hard to customize (by using a very specific selector).

app/assets/stylesheets/geoblacklight/record.css
```diff
@@ -14,7 +18,12 @@
 
 /* Attribute table */
 [data-sidebar=false] .attribute-data {
-  margin: 1rem 0;
+  margin-top: 1rem;
+}
+
+/* Metadata */
+.document-metadata {
+  margin-top: 1rem;
```
jcoyne, line 26: Is there a reason we aren't using `.mt-3` on the html element?  That is the general approach used by Blacklight.  It seems odd use a different approach for GBL.

thatbudakguy: I'm trying not to use bootstrap's helper classes because they apply `!important`, which imo is bad practice and makes it harder for people to override them in the generated application.

app/assets/stylesheets/geoblacklight/record.css
thatbudakguy: I guess it's not essential, but the spacing on the header seemed unbalanced without it:

Screenshot 2026-06-17 at 09 09 14

Screenshot 2026-06-17 at 09 07 52

Not sure what you mean by "hard to customize". The selector doesn't even include any IDs; you can override it pretty easily in your own app. I guess we could take out the first class, since maybe `.documentHeader h1` only occurs on the show page anyway?

jcoyne: Yes, `.documentHeader` is only on the show page.

thatbudakguy: I've shortened the selector (and the others that were using `.blacklight-catalog-show`).

jcoyne (approved):

PR #1816, opened 2026-06-17
Require geo_combine in the engine

### LINKED ISSUE #1798: Make GeoCombine rake tasks available in the built app
When a new application is built using `template.rb` or `engine_cart:generate`, it should immediately have access to the rake tasks provided by GeoCombine.

jcoyne:
This enables the geo_combine rake tasks to be loaded into the application.

Fixes #1798

### DIFF
--- a/lib/geoblacklight/engine.rb
+++ b/lib/geoblacklight/engine.rb
@@ -7,6 +7,7 @@ require "faraday"
 require "geoblacklight/version"
 require "nokogiri"
 require "mime/types"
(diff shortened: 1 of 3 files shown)


thatbudakguy (approved):
Tested this with:
`BRANCH=geo_combine rails new test-gbl -m https://raw.githubusercontent.com/geoblacklight/geoblacklight/refs/heads/main/template.rb`

And I saw the rake tasks appear in the built app. nice!

PR #1817, opened 2026-06-17
Nest css selectors

jcoyne:
This makes it clearer what the scope is


thatbudakguy (approved):
Didn't know this was built into CSS now!

PR #1821, opened 2026-06-18
Make the geosearch map have data-turbo-permanent

### LINKED ISSUE #1424: Use turbo frames to reload search results and pagination after geosearch
### Descriptive summary

Include what version of GeoBlacklight related to this issue (1.3.0, main, etc.) if appropriate, and any relevant tracebacks if you're reporting a bug.

### Expected behavior
Instead of reloading the entire page, just refresh the search results when geosearch occurs / the bounding box changes.

### Actual behavior
The whole page reloads

1. Zoom in on a map or pan the map when geosearch is enabled. The bounding box will chnage.
1. Notice the whole page reloads.

Essentially, turn the middle column of "1-4 of 4" and its content into a turbo frame.
Screenshot 2024-07-25 at 13 46 31

jcoyne:
So it doesn't flash

### DIFF
--- a/app/components/geoblacklight/location_leaflet_map_component.rb
+++ b/app/components/geoblacklight/location_leaflet_map_component.rb
@@ -35,19 +35,28 @@ module Geoblacklight
       Geoblacklight.configuration
     end
 
+    # We don't want turbo-permanent on the home page.
+    def data_permanent
+      return {} unless helpers.has_search_parameters?
+      {turbo_permanent: true}
+    end
+
     def viewer_tag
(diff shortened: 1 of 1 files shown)


thatbudakguy (changes requested):
This achieves what I was hoping for in #1424, but it has a weird side effect. If you run a no-text search from the homepage, you see this:

Image

And if you then return to the homepage, you see this:

Image

thatbudakguy (approved):
Looks like this is working great now!

PR #1830, opened 2026-06-25
Enhance LegacySettingsBuilder so defaults are used

jcoyne:
This allows us to trim away settings that are just copying the defaults. This makes it clearer when we've chosen to make a change from the defaults


thatbudakguy (approved):

thatbudakguy:
@jcoyne needs rebase

PR #1846, opened 2026-06-30
Drop Rails 7 support

jcoyne:
Rails 7 is EOL on 09 Aug 2026, so nobody should be using it for a new major version of GeoBlacklight

### DIFF
--- a/geoblacklight.gemspec
+++ b/geoblacklight.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = ">= 3.2.0"
   spec.required_rubygems_version = ">= 2.5.2"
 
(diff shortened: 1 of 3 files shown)


geoblacklight.gemspec
```diff
@@ -28,7 +28,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "mime-types"
   spec.add_dependency "rgeo-geojson"
   spec.add_dependency "rsolr"
-  spec.add_dependency "zeitwerk"
```
thatbudakguy, line 31: why does Zeitwerk go away?

jcoyne: It's a rails dependency already

thatbudakguy (approved):

PR #1859, opened 2026-07-08
Do not directly depend on rsolr

jcoyne:
It's more of a peer dependency generated into the gemfile by blacklight.

Reverts part of #1612


jcoyne:
Superseded by #1860

PR #1860, opened 2026-07-08
Remove rsolr dependency and injection step

thatbudakguy:
Blacklight already handles adding this to the built app's Gemfile
for us: https://github.com/projectblacklight/blacklight/blob/34d83482d341b51be40cf8fd07a3038de8c2ca1f/lib/generators/blacklight/solr_generator.rb#L33-L35

### DIFF
--- a/geoblacklight.gemspec
+++ b/geoblacklight.gemspec
@@ -28,7 +28,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "geo_combine", "~> 0.10"
   spec.add_dependency "mime-types"
   spec.add_dependency "rgeo-geojson"
-  spec.add_dependency "rsolr"
 
   spec.add_development_dependency "rails-controller-testing"
   spec.add_development_dependency "rspec-rails"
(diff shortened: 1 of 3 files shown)


jcoyne:
Duplicate of #1859?

jcoyne:
This goes a step further than #1859, so we can use this one, but I wanted to note this reverts part of https://github.com/geoblacklight/geoblacklight/pull/1612

thatbudakguy:
I think maybe we still need it as a development dependency?

jcoyne:
@thatbudakguy why do you think that?

thatbudakguy:
Because CI failed, looking for RSolr?

thatbudakguy:
See https://github.com/geoblacklight/geoblacklight/actions/runs/28956276954/job/85915976191

jcoyne:
@thatbudakguy it didn't fail on #1859.

thatbudakguy:
Hmm. Is the step that Blacklight runs to modify the Gemfile not running? it seems like we're not getting the dependency added in CI.

jcoyne:
In test_app_generator.rb we have:
```
    generate "blacklight:install", "--devise", "--skip-solr"

```

jcoyne (approved):

PR #1861, opened 2026-07-08
Fix truncator CSS so descriptions are partially shown

thatbudakguy:
We lost some CSS that applied to the truncated abstracts as part
of the move to pure CSS. As a result, long descriptions were fully
truncated to just the "read more" link, instead of a preview of a
few lines before the link.

This restores the needed CSS, and adds a small gradient effect to
make the bottom of the truncated area a little cleaner.

### DIFF
--- a/app/assets/stylesheets/geoblacklight/global.css
+++ b/app/assets/stylesheets/geoblacklight/global.css
@@ -5,3 +5,33 @@
   width: 1em;
   height: 1em;
 }
+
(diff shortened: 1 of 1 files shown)


thatbudakguy:
# Before
 
Screenshot 2026-07-08 at 09 52 58

# After

Screenshot 2026-07-08 at 09 49 33

jcoyne (approved):

PR #1881, opened 2026-07-16
Reformat everything with prettier

thatbudakguy:
Now you can `npx prettier --write .` without a whole bunch of other stuff cropping up.


jcoyne (commented):
I think it would be a good idea to put the format validator in the CI too.

thatbudakguy:
It is: https://github.com/geoblacklight/geoblacklight/blob/9abdf61bcbd6f832631d05012162ab1f533f52dd/.github/workflows/ruby.yml#L34

It's just restricted to some file types: https://github.com/geoblacklight/geoblacklight/blob/9abdf61bcbd6f832631d05012162ab1f533f52dd/package.json#L12

jcoyne (approved):

jcoyne (approved):

PR #1890, opened 2026-07-21
Convert icons into badges

### LINKED ISSUE #1849: Display metadata as list of badges on search result view
This is an experiment in updating the search result list view. Right now we have a generated text field hidden behind a disclosure widget that can be expanded; the text contains the values of selected metadata fields like the date and description. We use icons to indicate the values of some other fields (resource class/type, access rights):

Image

What if instead, we created a new metadata display that rendered the values (but not field names) of a selected set of fields as bootstrap "badges"? This would result in a "tag"-like appearance, e.g.

Image

I think this might be an interesting compromise between the icons (a little cryptic, take up space in the title) and the generated description (needs to be expanded, appears kind of weird for some records).

jcoyne:
Fixes #1849
Screenshot 2026-07-21 at 2 47 29 PM

### DIFF
--- a/app/components/geoblacklight/header_badge_component.html.erb
+++ b/app/components/geoblacklight/header_badge_component.html.erb
@@ -1,0 +1,2 @@
+<span class="badge text-bg-secondary text-uppercase fw-semibold d-inline-flex align-items-center gap-1 me-1">
+  <%= label %><%= icon(field) %></span>
--- a/app/components/geoblacklight/header_badge_component.rb
+++ b/app/components/geoblacklight/header_badge_component.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class HeaderBadgeComponent < ViewComponent::Base
+    attr_reader :document, :field
+
+    def initialize(document:, field:)
+      @document = document
+      @field = field
+      super()
+    end
+
+    def label
+      @document[field]
+    end
+
+    def icon(field)
+      helpers.geoblacklight_icon(label)
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


app/components/geoblacklight/header_badge_component.html.erb
```diff
@@ -0,0 +1,2 @@
+<span class="badge text-bg-secondary text-uppercase fw-semibold d-inline-flex align-items-center gap-1 me-1">
```
thatbudakguy, line 1: Can we do these as styles in geoblacklight's stylesheet so they don't apply `!important`?

jcoyne: The supported way to do that is to include a custom. version of bootstrap without the important flag.  Otherwise we're basically re-implementing a big chunk of bootstrap.

thatbudakguy: I just want a class name on this that applies those styles – not a generic class to apply everywhere. We don't want to be compiling our own bootstrap; that's why we got rid of the SCSS pipeline.

jcoyne: We got rid of the SCSS pipeline because this is a library, but in your own app you can do it if you want.  Using utility classes is the way blacklight does this, so since this extends from Blacklight we probably want to follow the same approach, so that the fix is the same for both. 

If any one is compiling their own SCSS, they will expect the utility classes to adapt when they change settings like $spacer.  If we write our own css classes, they will not adapt to those changes.

thatbudakguy (approved):

PR #1900, opened 2026-07-22
Convert the location map to a facet

### LINKED ISSUE #1889: Move geosearch map to the facet area
What if we treated the map that appears on the search results page as a facet? This would allow us to collapse to a two-column layout matching other parts of the site (instead of 3, which is a little cramped) and keep all of the "search modification tools" in the facets column.

I'm imagining a small version of the map that just indicates where the results on the page are located, and maybe allows you to execute a geosearch as well. It would appear at the top of the facets list.

thatbudakguy:
Closes #1889

### DIFF
--- a/app/components/geoblacklight/location_leaflet_map_component.rb
+++ b/app/components/geoblacklight/location_leaflet_map_component.rb
@@ -3,14 +3,10 @@
 module Geoblacklight
   class LocationLeafletMapComponent < ViewComponent::Base
     def initialize(
-      id: "leaflet-viewer",
       map_geometry: config.homepage_map_geom,
-      classes: "",
       page: nil,
       geosearch: nil
     )
-      @id = id
-      @classes = classes
       @map_geometry = map_geometry if map_geometry != "null"
       @page = page
       @geosearch = geosearch
@@ -49,9 +45,11 @@ module Geoblacklight
     end
 
     def viewer_tag
-      tag.div(id: @id,
-        class: @classes,
-        data: data_permanent.merge(leaflet_viewer_data_attributes))
+      tag.div(
+        id: "leaflet-viewer",
+        class: "viewer leaflet-viewer location-viewer",
+        data: data_permanent.merge(leaflet_viewer_data_attributes)
+      )
     end
 
     def leaflet_viewer_data_attributes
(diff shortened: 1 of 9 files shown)


app/components/geoblacklight/location_leaflet_map_component.rb
```diff
@@ -61,7 +59,7 @@ def leaflet_viewer_data_attributes
         "leaflet-viewer-dark-basemap-value" => Geoblacklight.configuration.dark_basemap_provider,
         "leaflet-viewer-map-geom-value" => search_bbox || @map_geometry,
         "leaflet-viewer-data-map-value" => @data_map,
-        "leaflet-viewer-page-value" => params[:action]&.upcase,
+        "leaflet-viewer-page-value" => @page || params[:action],
```
thatbudakguy, line 62: this resulted in values like `INDEX`, but our config actually has them lowercased. on the homepage, `params[:action]` is empty, so we want the passed `home` value to take precedence.

lib/generators/geoblacklight/templates/catalog_controller.rb
```diff
@@ -86,35 +87,36 @@ class CatalogController < ApplicationController
...
+    config.add_facet_field field_config.geometry, component: Geoblacklight::Facets::BboxFieldComponent,
+      item_presenter: Geoblacklight::BboxItemPresenter,
       filter_class: Geoblacklight::BboxFilterField,
       filter_query_builder: Geoblacklight::BboxFilterQuery,
       within_boost: gbl_config.bbox_within_boost,
       overlap_boost: gbl_config.overlap_ratio_boost,
       overlap_field: field_config.overlap_field,
       label: "Bounding Box"
 
+    # DEFAULT FACETS
```
thatbudakguy, line 106: Moved these down so the bbox facet comes first.

thatbudakguy:
# Screenshots

## Big
Screenshot 2026-07-22 at 15 38 52

## Medium
Screenshot 2026-07-22 at 15 38 59

## Mobile
Screenshot 2026-07-22 at 15 40 02

jcoyne:
on mobile are the facets open by default?

thatbudakguy:
No, they are closed on mobile until you click "show facets"

jcoyne (approved):

PR #1902, opened 2026-07-23
Add instructions for installing blacklight_date_range

### LINKED ISSUE #1899: Add blacklight_range_limit for date display
It's a much better option than the default "list of years" presentation – enough that it probably deserves to be part of core.

We should also consider porting the UI changes in [DataWorks's version](https://github.com/sul-dlss/DWExperimentsBlacklight/blob/main/app/components/index/range_facet_component.html.erb), either to here or to blacklight_range_limit directly.

jcoyne:
Fixes #1899

### DIFF
--- a/lib/generators/geoblacklight/templates/catalog_controller.rb
+++ b/lib/generators/geoblacklight/templates/catalog_controller.rb
@@ -100,6 +100,10 @@ class CatalogController < ApplicationController
     config.add_facet_field field_config.provider, label: "Provider", limit: 8
     config.add_facet_field field_config.georeferenced, label: "Georeferenced", limit: 3
 
+    # If you wish to add a date range filter, first run `bundle add blacklight_range_limit && bin/rails generate blacklight_range_limit:install`
+    # Then uncomment the following line:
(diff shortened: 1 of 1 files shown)


lib/generators/geoblacklight/templates/catalog_controller.rb
```diff
@@ -100,6 +100,10 @@ class CatalogController < ApplicationController
     config.add_facet_field field_config.provider, label: "Provider", limit: 8
     config.add_facet_field field_config.georeferenced, label: "Georeferenced", limit: 3
 
+    # If you wish to add a date range filter, first run `bundle add blacklight_range_limit && bin/rails generate blacklight_range_limit:install`
```
thatbudakguy, line 103: Does this also handle the dependencies part? I'm wondering if the pin we had to do in Earthworks needs to be part of these instructions or if that's what the install generator does

jcoyne: yes, the generator does everything.

thatbudakguy (approved):

PR #1903, opened 2026-07-23
Fix download link badge wrapping for long titles

thatbudakguy:
Items with download links that are URLs or filenames didn't wrap
very well insdie the download card. The float-based solution
worked, but not on Firefox. This fixes it for all browsers.


thatbudakguy:
# Long text
Screenshot 2026-07-23 at 16 08 49

# Long filenames
Screenshot 2026-07-23 at 16 08 55

jcoyne (approved):

PR #1905, opened 2026-07-24
Update stanford fixtures and related tests

thatbudakguy:
- **Replace restricted-line fixture**
  This piece of Stanford data is a COG now, so it isn't as good
  for testing what it was for.
  
  Replaces it with a piece of restricted line data from Berkeley
  that is available over WMS/WFS, which is pretty close to what
  the data was originally.
  

- **Remove redundant tests**
  These tests were basically just asserting that basic Blacklight
  behavior works correctly. That's already tested for us by Blacklight.
  

- **Update Stanford fixtures**
  These items have different links to metadata and data now, since
  we no longer have a geoserver. In the case of the index map,
  switched it to point at a different one that's available, because
  the old one no longer is.
  
  Also updates the metadata modal spec to consolidate the tests
  under one item with all the relevant metadata types.


jcoyne (changes requested):
It seems that we're losing tests that check javascript functionality and not replacing them with anything.  Do we really want to not check those things?

thatbudakguy:
Which tests are you worried about?

jcoyne:
* [metadata_panel_spec.rb](https://github.com/geoblacklight/geoblacklight/pull/1905/changes#diff-31cb9d7034a790bfcbff9469d4b9e9462874a8a5193e7eab0684bb495b941a5b)
* [sms_spec.rb](https://github.com/geoblacklight/geoblacklight/pull/1905/changes#diff-9ee6c49e23a948c3d1108b8144311a911af6525981ac7f0bfa7c5ed06f9d2e36)

thatbudakguy:
I see what you mean re: the metadata panel (the request spec tests it, but not the modal opening part). I've restored that and updated it to use the newer fixtures.

SMS I thought was fully standard Blacklight functionality that we hadn't modified at all – is that not true?

jcoyne:
It seems it was added intentionally here: https://github.com/geoblacklight/geoblacklight/commit/03f2eb5f27ff5dc0f1e713c5e66192ac6df20695

jcoyne (approved):

PR #1906, opened 2026-07-30
Add and use rake binstub

thatbudakguy:


jcoyne (approved):

thatbudakguy:
@jcoyne do you know why this isn't working? the error messages are confusing.

jcoyne:
I think `bin/rake` within geoblacklight is conflicting with `exe/rake` inside the rake gem.  Typically binstubs only work in Rails applications, no within gems.

PR #1912, opened 2026-08-26
Fix the issue where the metadata modal button redirects to the original page - #1911

### LINKED ISSUE #1910: Metadata modal button redirects to original page instead of opening XML metadata in a new browser window ( Version: 5.3.0 and current main)
When clicking the button in metadata modal, it redirects to the original page instead of displaying the XML metadata content in a new window.

This issue occurs in GeoBlacklight 5.3.0 and is also reproducible on the current main branch of the GeoBlacklight repository.

yzhoubk:
1. Use the Blacklight modal to dispatch custom events within the metadata modal
2. Add `aria-label` to the metadata download link to improve accessibility

### DIFF
--- a/app/javascript/geoblacklight/initializers/metadata_download.js
+++ b/app/javascript/geoblacklight/initializers/metadata_download.js
@@ -35,7 +35,6 @@ export default function initializeMetadataDownload() {
     })
 
     e.target.querySelectorAll(".pill-metadata").forEach((element, i) => {
-      console.log(element)
       new GeoBlacklightMetadataDownloadButton(element, i)
     })
   })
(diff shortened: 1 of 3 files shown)


app/javascript/geoblacklight/initializers/metadata_download.js
```diff
@@ -28,23 +28,19 @@ export class GeoBlacklightMetadataDownloadButton {
...
-      return
-    }
+  // Use Blacklight's dialog-based modal dispatches these custom events
+  modal.addEventListener("show.blacklight.blacklight-modal", (e) => {
     e.target.querySelectorAll(".metadata-body").forEach((el) => {
       el.closest(".modal-content").classList.add("metadata-modal")
     })
 
     e.target.querySelectorAll(".pill-metadata").forEach((element, i) => {
+      console.log(element)
```
thatbudakguy: Left over from debugging?

yzhoubk: Yes, thanks for catching that. I've removed it.

thatbudakguy (approved):
Thanks for fixing this!

PR #1914, opened 2026-08-26
Add configurable thumbnails to search results

### LINKED ISSUE #1898: Render configurable thumbnails in index view
This should be able to be enabled/disabled via a config value.

A good field to use (though it's not in the Aardvark spec) is a `dct_references_s` entry with key `http://schema.org/thumbnailUrl`. We have been populating this locally at Stanford.

A good UI model to draw from is the [big ten catalog](https://geo.btaa.org/search?q=) (though no longer a geoblacklight). When a thumbnail isn't available, they show a placeholder icon depending on the data type.

If there is existing logic to use/borrow from Blacklight here, we should do that.

thatbudakguy:
Renders a thumbnail per document in search results, sourced from a
configurable dct_references_s key (defaulting to
http://schema.org/thumbnailUrl) and falling back to a resource-type
icon when no image is available. Reuses Blacklight's existing
thumbnail_method/default_thumbnail/ThumbnailComponent machinery
rather than adding new rendering infrastructure.

Closes #1898

### DIFF
--- a/app/assets/stylesheets/geoblacklight/search.css
+++ b/app/assets/stylesheets/geoblacklight/search.css
@@ -7,6 +7,43 @@
   .document {
     border: 0;
     padding-bottom: var(--bl-results-document-padding-top);
+
+    .document-thumbnail {
+      padding: 0;
+      margin: 0 1rem 0 0;
+
+      a {
+        color: var(--bs-body-color);
+        background-color: var(--bs-body-bg);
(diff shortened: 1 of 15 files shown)


thatbudakguy:
Screenshot 2026-08-26 at 09 33 44

jcoyne (approved):

PR #1920, opened 2026-08-28
Derive a thumbnail URL from the IIIF info.json URL where possible

thatbudakguy:
This adds some logic for generating a thumbnail URL when we don't
have one, but we do have an info.json URL.


dl-maura (approved):
See a thumbnail locally for the test item

PR #1927, opened 2026-08-31
More fixture fixes

thatbudakguy:
- **Remove Harvard fixture**
  Harvard is about to transition to cloud-only fixtures and will be
  retiring their geoserver, per @dl-maura.
  
  This record was only used as a test for items with no relationships,
  so this switches those tests to use a different fixture instead.
  

- **Remove dead UVA fixture**
  This fixture wasn't used to test anything and didn't resolve. It's
  from an organization that no longer participates in OGM.
  

- **Update the fixtures README**


thatbudakguy:
This handles all of the documents named in #1492 except the NYU ones – I'm hoping we can get them to turn on CORS on their geoserver.

jcoyne (approved):

PR #1936, opened 2026-09-01
Badge updates; add a georeferenced badge

thatbudakguy:
Changes:
- Rename HeaderIconsComponent to HeaderBadgesComponent
- Move badge styles to CSS; use a gentler appearance
- Add a badge for georeferenced records

After this, badges are a little more muted, in keeping with the rest of the page's style. They're also styled without using helper classes anymore, so they're easier to override in downstream apps.

Screenshot 2026-09-01 at 14 11 55

You also get one for georeferenced items.

Screenshot 2026-09-01 at 14 12 02


app/assets/stylesheets/geoblacklight/global.css
```diff
@@ -32,3 +32,14 @@
     background-image: linear-gradient(to bottom, rgba(255, 255, 255, 0), var(--bs-body-bg) 65%);
   }
 }
+
+/* Badges */
+.badge {
```
jcoyne: Is this intended to replace the bootstrap badge style?

https://getbootstrap.com/docs/5.3/components/badge/#examples

Some of these choices like "text-transform: uppercase" seem rather heavy handed and I wonder if they should be left to the host app to do. 

How about using `.d-inline-flex` to the element rather than making every `.badge` in the app have this alignment? 

I believe there is already a `--bs-badge-color` variable that should be used. 

The supported way of setting a background color on a badge is to use `.text-bg-secondary`

app/components/geoblacklight/header_badge_component.html.erb
```diff
@@ -1,2 +1,4 @@
-<span class="badge text-bg-secondary text-uppercase fw-semibold d-inline-flex align-items-center gap-1 me-1">
-  <%= label %><%= icon(field) %></span>
+<span class="badge">
+  <%= tag.span label, class: 'label' %>
```
jcoyne, line 2: are we relying on this `.label` class for anything?

app/components/geoblacklight/header_badges_component.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Geoblacklight
+  class HeaderBadgesComponent < ViewComponent::Base
+    attr_reader :document, :fields
```
jcoyne: Why do we have an attr_reader for `fields`?

jcoyne (changes requested):

app/assets/stylesheets/geoblacklight/global.css
thatbudakguy: these are just replacements for each of the helper classes we used to apply to the badge in the template that used `!important`, so that the built app can change them if it wants to.

app/components/geoblacklight/header_badge_component.html.erb
thatbudakguy: nope, just wanted to distinguish the text span from the icon span.

app/components/geoblacklight/header_badges_component.rb
thatbudakguy: dunno! removed it.

app/assets/stylesheets/geoblacklight/global.css
jcoyne: What if the built app uses .badge in other places?  This makes it harder to get back to the bootstrap baseline, doesn't it?  Instead can we make it easy to configure/replace the component that sets the classes so you can just not set them if you don't like them?

thatbudakguy: OK – moved these to two unique classes that apply to the two places we make badges.

jcoyne (approved):

PR #1971, opened 2026-09-17
Provide an icon for the "Other" resource class

thatbudakguy:
This moves upstream a local customization we had at Stanford that
provided this icon, since it was the only resource class without one.

Now items with this resource class are guaranteed something to
display as a thumbnail; before they could be totally blank.


thatbudakguy:
What it looks like:

Screenshot 2026-09-17 at 08 58 26

